PR #2, opened 2020-03-06
Alternative view for widgets with no data

denniskigen:
This PR introduces an alternative view which is rendered when widgets do not have data to display.

Presently, most widgets will render an incomplete view - in the case of the Notes widget, table headers will be displayed without the corresponding table data.

Screenshot 2020-03-03 at 21 04 11
 
This could be improved upon by rendering something that communicates that the patient does not have recorded patient notes and perhaps a link to a view where patient notes could be added.

Screenshot 2020-03-03 at 21 13 29

This is how the patient chart would look if all the widgets did not have data:

Screenshot 2020-03-04 at 12 15 12

### DIFF
--- a/__mocks__/allergy.mock.ts
+++ b/__mocks__/allergy.mock.ts
@@ -1558,3 +1558,96 @@ export const mockAllergyResult = {
     }
   }
 };
+
+export const mockPatientAllergiesResult = {
+  data: {
+    total: 2,
+    entry: [
+      {
+        resource: {
+          resourceType: "AllergyIntolerance",
+          id: "67a8dad8-0d35-4afd-a838-f96913614c53",
+          category: ["medication"],
+          criticality: "?",
+          code: {
+            coding: [
+              {
+                system: "http://openmrs.org",
+                code: "162298AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
+                display: "ACE inhibitors"
+              }
+            ],
+            text: "ACE inhibitors"
+          },
+          note: [
+            {
+              text: "comment"
+            }
+          ],
+          reaction: [
+            {
+              manifestation: [
+                {
+                  coding: [
+                    {
+                      system: "http://openmrs.org",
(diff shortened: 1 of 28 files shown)


src/widgets/medications/medications-overview.component.tsx
```diff
@@ -202,6 +173,43 @@ export default function MedicationsOverview(props: MedicationsOverviewProps) {
       );
     });
   }
+
+  function displayMedications() {
```
nkimaina, line 177: **For all places where you are using this pattern**
1. Main component vs supporting component: It seems to be the norm in react that the JSX bit of the component goes at the end of the function component as a return statement. The supporting components(which is consumed by the main component) can be declare inline, or in separate files. The component in question here is the 'no content' component which appears to be the main component of the function. 
2. Use of lambda expressions to avoid running into bugs: you can avoid the risk of bugs due to 'this' reference by using lambda expressions as:
`const noMedications = () => (<><>)`
3 You are repeating yourself a lot. Maybe that indicates that you need to refactor the 'no content' component into a hook/function component?

__mocks__/allergy.mock.ts
```diff
@@ -1558,3 +1558,96 @@ export const mockAllergyResult = {
     }
```
nkimaina, line 1558: Generally: To avoid the PR from being intimidating and therefore easier to review, you need to make it smaller by breaking it down. You could have split the PR into a) refactors b) The main purpose of the PR, which is to display something when there's not records for a particular component.

nkimaina (changes requested):

__mocks__/allergy.mock.ts
denniskigen: Thanks, @nkimaina for the feedback. I'm closing this PR in favour of #44 and https://issues.openmrs.org/projects/MF/issues/MF-182?filter=allissues where I'll execute this better.

PR #4, opened 2020-03-09
Create meds app

jonathandick:
This creates a Medications widget which consists of the level 2 (detailed summary) and level 3 (details for an individual medication order) components. 
The Medications Overview widget will be treated as a separate widget in the api. The overview component takes a property called "basePath" which it uses internally to build links to the primary Medications widget.


src/widgets/medications/medications.component.tsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import { Switch, Route, useRouteMatch } from "react-router-dom";
+import MedicationsSummary from "./medication-level-two.component";
+import MedicationDetailedSummary from "./medication-level-three/medication-level-three.component";
+
+function Medications(props) {
+  const match = useRouteMatch();
+
+  return (
+    <Switch>
```
fatmali, line 10: 👏 nice breaking the routes down like this. Makes alot of sense

fatmali (approved):

PR #6, opened 2020-03-10
Enhancements to the medications widget

denniskigen:
This PR enhances the medications widget in the following ways:

- Adds a HorizontalLabelValue with the text 'Active Medications' to the MedicationsOverview component.
- Adds visual distinction to the Revise and Discontinue medication action buttons by applying different colours.
- Fixes issue where some missing properties on the medication API response would cause the UI to error out.
- Fixes the link from the MedicationsOverview component in the summary chart so that it now navigates you to the MedicationsDetailedSummary component in the Orders tab.
- Introduces an alternative view to the MedicationsDetailedSummary component which is rendered when the patient does not have any active or past medications.
- Adds a bunch of UI and styling enhancements to the MedicationsDetailedSummary component from the figma designs.
- Adds a bunch of UI and styling enhancements to the MedicationRecord component from the figma designs.
- Modifies the components in the Medication Order widget to use the new navigation and file structure.
- Adds the ability to modify an existing medication order by launching the MedicationOrderBasket form from the MedicationRecord component.

Screenshot 2020-03-10 at 16 30 55

Screenshot 2020-03-10 at 16 32 18

Screenshot 2020-03-10 at 22 55 05

Screenshot 2020-03-23 at 15 38 23

Screenshot 2020-03-23 at 17 33 08


jonathandick:
@denniskigen , thanks for this work
1. These are medication orders not a medication history. 
2. We should remove the footer that says no more active medications available. This is implied as this list is meant to be complete.

src/widgets/medications/medication-level-two.test.tsx
```diff
@@ -1,65 +1,119 @@
...
+  fetchPatientMedications: jest.fn()
+}));
+
+jest.mock("@openmrs/esm-api", () => ({
+  useCurrentPatient: jest.fn()
+}));
+
+let wrapper;
+
+describe("<MedicationLevelTwo/>", () => {
```
jonathandick: could we use new naming conventions instead of MedicationLevelTwo?

jonathandick (changes requested):

src/widgets/medications/medication-level-two.test.tsx
denniskigen: Sure

denniskigen:
> @denniskigen , thanks for this work
> 
> 1. These are medication orders not a medication history.
> 2. We should remove the footer that says no more active medications available. This is implied as this list is meant to be complete.

Noted. I'll make the necessary changes.

jonathandick (approved):

PR #8, opened 2020-03-11
Fix ProgramsSummary widget import

denniskigen:
Fix the ProgramsSummary widget export so that it exports the ProgramsSummaryComponent rather than the obsolete ProgramsLevelTwo component.

I want to use the ProgramsSummary component while working on adding the ability to enroll or unenroll from programs.


jonathandick (approved):

tendomart (commented):
I find it clear.

PR #12, opened 2020-03-18
MF-141: Update Programs Navigation and Structure

denniskigen:
This PR modifies the programs widget so that it corresponds with the routing and navigation structure specified in https://issues.openmrs.org/projects/MF/issues/MF-141.

Screenshot 2020-03-20 at 13 07 05

Screenshot 2020-03-20 at 13 07 14


jonathandick:
This is good work @denniskigen.

For now, please don't update configs for views. The conditions tab should be remain the same.

src/widgets/programs/programs-overview.component.tsx
```diff
@@ -35,7 +39,7 @@ export default function Programs(props: ProgramsOverviewProps) {
   return (
     <SummaryCard
       name={t("care programs", "Care Programs")}
-      link={`/patient/${patientUuid}/chart/programs`}
+      link={`${props.basePath}`}
```
jonathandick: This should be programsPath not props.basePath

jonathandick (changes requested):
One small change and then we can merge.

src/widgets/programs/programs-overview.component.tsx
denniskigen: Fixed

denniskigen:
> This is good work @denniskigen.
> 
> For now, please don't update configs for views. The conditions tab should be remain the same.

Ok, this was only for the purposes of getting these screenshots

PR #15, opened 2020-03-18
MF-140: Update Conditions Navigation and Structure

denniskigen:
This PR modifies the conditions widget so that it corresponds with the routing and navigation structure specified in https://issues.openmrs.org/projects/MF/issues/MF-140.

Screenshot 2020-03-20 at 12 58 22

Screenshot 2020-03-20 at 12 48 00


src/widgets/conditions/conditions-overview.component.tsx
```diff
@@ -39,6 +43,7 @@ export default function ConditionsOverview(props: ConditionsOverviewProps) {
     <SummaryCard
       name={t("conditions", "Conditions")}
       styles={{ margin: "1.25rem, 1.5rem" }}
+      link={`${props.basePath}`}
```
jonathandick: This should be conditionsPath

jonathandick (changes requested):
One small change and then we can merge.

jonathandick (approved):

PR #20, opened 2020-03-23
Follow-up fixes for the Conditions widget

denniskigen:
Fixes that should have been in #15 :

- Render the selected condition rather than rendering the same one regardless of which condition was selected.
- Change the condition link in ConditionsOverviewComponent to pick the `resource.id` from the selected condition - there is no `uuid` property.
- Change the glob in the `ConditionRecord` route path in the Conditions component to match `conditionUuid` instead of `programUuid`.
- Change `ConditionsSummary` export to `Conditions`
- Got rid of `No more conditions available` text in SummaryCard footer


donaldkibet (approved):
lgtm

jonathandick (approved):

PR #32, opened 2020-04-01
MF-130: Implement a form for adding and editing conditions

denniskigen:
Re-opened from https://github.com/openmrs/openmrs-esm-patient-chart/pull/97.

A reminder that POST and DELETE won't work on this form yet - the Conditions API isn't at the stage yet where these features can be fully implemented. This will be rectified once that API is ready.

Screenshot 2020-04-01 at 19 16 56
Screenshot 2020-04-01 at 19 18 38

### DIFF
--- a/src/widgets/conditions/condition-record.component.tsx
+++ b/src/widgets/conditions/condition-record.component.tsx
@@ -1,26 +1,54 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 import { useRouteMatch } from "react-router-dom";
 import { getConditionByUuid } from "./conditions.resource";
 import { createErrorHandler } from "@openmrs/esm-error-handling";
+import { useCurrentPatient, newWorkspaceItem } from "@openmrs/esm-api";
 import dayjs from "dayjs";
 import styles from "./condition-record.css";
 import SummaryCard from "../../ui-components/cards/summary-card.component";
-import { useCurrentPatient } from "@openmrs/esm-api";
+import { ConditionsForm } from "./conditions-form.component";
 
 export default function ConditionRecord(props: ConditionRecordProps) {
-  const [patientCondition, setPatientCondition] = React.useState(null);
+  const [patientCondition, setPatientCondition] = useState(null);
   const [isLoadingPatient, patient, patientUuid] = useCurrentPatient();
   const match = useRouteMatch();
 
-  React.useEffect(() => {
+  useEffect(() => {
     if (!isLoadingPatient && patient) {
       const sub = getConditionByUuid(match.params["conditionUuid"]).subscribe(
-        condition => setPatientCondition(condition),
+        ({ resource }) => setPatientCondition(resource),
         createErrorHandler()
       );
       return () => sub.unsubscribe();
     }
   }, [isLoadingPatient, patient, match.params]);
+
+  const openEditConditionsWorkspaceTab = (
+    componentName,
+    title,
+    conditionUuid,
+    name,
+    clinicalStatus,
+    onsetDateTime
+  ) => {
+    newWorkspaceItem({
+      component: componentName,
+      name: title,
+      props: {
+        match: {
+          params: {
+            conditionUuid,
+            name,
+            clinicalStatus,
+            onsetDateTime
+          }
+        }
+      },
+      inProgress: false,
+      validations: (workspaceTabs: any[]) =>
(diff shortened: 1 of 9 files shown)


src/widgets/conditions/condition-record.component.tsx
```diff
@@ -1,39 +1,77 @@
...
+          params: {
+            conditionUuid,
+            name,
+            clinicalStatus,
+            onsetDateTime
+          }
+        }
+      },
+      inProgress: false,
+      validations: (workspaceTabs: any[]) =>
```
FlorianRappl: Why `any`? Do we have a type here? If you expect the `component` prop to be there then at least type it as `{ component: string }` to communicate that requirement.

src/widgets/conditions/condition-record.component.tsx
```diff
@@ -1,39 +1,77 @@
...
+              patientCondition.clinicalStatus,
+              patientCondition.onsetDateTime
+            );
+          }}
         >
           <div className={`omrs-type-body-regular ${styles.conditionCard}`}>
             <div>
               <p className="omrs-type-title-3" data-testid="condition-name">
-                {patientCondition.resource.code.text}
+                {patientCondition.code.text}
```
FlorianRappl: I am surprised here. Below we use `patientCondition?.` implying that `patientCondition` may be undefined - now we access `code.text` here without the conditional member operator. Is this desired?

Should it be `patientCondition?.code.text` instead?

src/widgets/conditions/conditions-utils.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+export const openConditionsWorkspaceTab = (componentToAdd, componentName) => {
+  newWorkspaceItem({
+    component: componentToAdd,
+    name: componentName,
+    props: {
+      match: { params: {} }
+    },
+    inProgress: false,
+    validations: (workspaceTabs: any[]) =>
```
FlorianRappl: Same as above - I would love to see `any[]` being replaced with a more specific type, or otherwise `Array<{ component: string }>`

FlorianRappl (changes requested):

src/widgets/conditions/condition-record.component.tsx
denniskigen: Given that it workspace tabs seems to expect an array of function components, do you think the appropriate type would be `Array<{ component: React.FunctionComponent }>`?

This pattern where widgets have their own specific tab launcher functions has been repeated in many places across the app - here Conditions has its own variant - `openConditionsWorkspaceTab`, Allergies has `openAllergiesWorkspaceTab` and so on. It's probably a good candidate for DRYing up.

FlorianRappl: So `componentName` refers to a  `React.FC`? Sure - I would also recommend renaming `componentName` to `component` to reflect this. Usually, a name is a string (at least to me).

Rest I fully agree! :+1:

src/widgets/conditions/condition-record.component.tsx
denniskigen: Sure. This was an oversight on my part. Corrected.

src/widgets/conditions/condition-record.component.tsx
denniskigen: No, `component` refers to a `React.FC`. `componentName` remains a string - the name of the component.

```
// workspace.resource.tsx (from openmrs-esm-api)
export function newWorkspaceItem(item: WorkspaceItem) {
  workspaceItem.next(item);
}

export type WorkspaceItem = {
  component: any;
  name: string;
  props: any;
  validations?: Function;
  inProgress: boolean;
  componentClosed?: Function;
};
```

denniskigen: > Given that it workspace tabs seems to expect an array of function components, do you think the appropriate type would be `Array<{ component: React.FunctionComponent }>`?
> 
> This pattern where widgets have their own specific tab launcher functions has been repeated in many places across the app - here Conditions has its own variant - `openConditionsWorkspaceTab`, Allergies has `openAllergiesWorkspaceTab` and so on. It's probably a good candidate for DRYing up.

#47 addresses the issue of making the newWorkspaceItem logic reusable.

FlorianRappl (approved):
LGTM!

PR #35, opened 2020-04-03
Fhir2

nkimaina:


jonathandick (commented):
To avoid this, might be better to have a global property which stores the url of whatever backend api we use (so we don't have to continually change this)?

nkimaina:
@jonathandick I made the requested changes. Please let me know if it's ready for merging so that we can deploy the new module to openmrs-spa.org

nkimaina:
The module fails CI process because it requires https://github.com/openmrs/openmrs-esm-api/pull/32 to be released.

donaldkibet (approved):

jonathandick (approved):

PR #37, opened 2020-04-08
MF-164: Preferred patient identifier check not working in fhir2

denniskigen:
https://issues.openmrs.org/browse/MF-164


kipchubett:
According to FHIR specification [here](https://www.hl7.org/fhir/valueset-identifier-use.html) identifier-use can have different values other than `official or usual`. Although, `official`  is the preferred/primary/main identifier-use. Still, I think this should configurable rather than hard-coded.

PR #39, opened 2020-04-14
MF-150: Add gallery component

ivange94:
This is a work in progress. This is a gallery widget component for the attachments module.


src/widgets/attachments/attachments-gallery.component.tsx
```diff
@@ -0,0 +1,126 @@
...
+  function handleDelete(id: number) {
+    const docs_new = documents.slice();
+    docs_new.splice(id, 1);
+    setDocuments(docs_new);
+  }
+
+  function handleUpload(e: React.SyntheticEvent, files: FileList | null) {
+    e.preventDefault();
+    e.stopPropagation();
+    if (files != null) {
```
FlorianRappl: Either use `!==` or just use `if (files)`.

src/widgets/attachments/attachments-overview.component.tsx
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+import AttachmentsGallery from './attachments-gallery.component';
+
+
+export default function AttachmentsOverview() {
```
FlorianRappl, line 11: Hm if you just want to re-export it as default then why not `export default AttachmentsGallery`? Any reason for the additional wrapping?

src/widgets/attachments/attachments-gallery.css
```diff
@@ -0,0 +1,14 @@
+.documentsContainer {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+input[type="file"] {
```
FlorianRappl: Not sure how this one gets scoped. My gut feeling is, that it is not. Hence please introduce a class to avoid clashes with other file inputs in the whole app.

src/widgets/attachments/attachment-document.component.tsx
```diff
@@ -0,0 +1,32 @@
...
+        </div>
+      );
+}
+
+type AttachmentDocumentProps = {
+    src: string;
+    caption?: string;
+    type: string;
+    id: number;
+    onDelete?: any;
```
FlorianRappl: Also don't forget to type these.

src/widgets/attachments/attachments-overview.component.tsx
ivange94: Actually `AttachmentGallery` was created after refactoring. Before all components where in AttachmentsOverview. I also felt I just needed to just export `AttachmentGallery` but at the same time I wasn't really sure of OpenMRS standards as I saw a lot of overview components as the the top level component.

src/widgets/attachments/attachments-gallery.css
ivange94: Thanks. Will fix.

src/widgets/attachments/attachments-gallery.component.tsx
ivange94: Thanks for catching that.

src/widgets/attachments/attachment-document.component.tsx
ivange94: for the function, I wasn't really sure what type it was supposed to be.

FlorianRappl: Good point. Thought from the usage it's something like `onDelete?(id: string): void;`. Note that the TypeScript code says its optional, but the prop/function is just called.

src/widgets/attachments/attachments-overview.component.tsx
FlorianRappl: Thanks for the clarification!

Makes sense - I am myself too unfamiliar with what is expected, so somebody with more historic knowledge here could comment better.

src/widgets/attachments/attachment-document.component.tsx
ivange94: Thanks. I can't actually remember why I made optional. I remember having an error at one point but can't remember  the error. But I've removed the optional from `onDelete` and `caption` as they no longer need it. But for the onDelete, I left the signature as `onDelete(id: number): void;`. Before this PR is merged I'll make the ID a string. The reason it's a number now is that I temporarily initialised the elements in a array so I'm using their index to access them. Currently working on add calls to the backend where the state will be set by data from the backend and I won't need to keep track of the elements with their index.

src/widgets/attachments/attachments-gallery.css
ivange94: Is this better?

```typescript
.galleryHeader input[type="file"] {
  display: none;
}
```

src/widgets/attachments/attachments-overview.component.tsx
ivange94: It now makes sense to leave `AttachmentOverview` as a separate component. I have refactored the code and lifted `attachments` state from `AttachmentGallery` to `AttachmentOverview`. And AttachmentOverview just uses 

```html
<AttachmentsGallery
      attachments={attachments}
      onDelete={handleDelete}
      onAdd={handleAdd}
    />
```

This way it becomes easier to swap `AttachmentGallery` with another implementation.

FlorianRappl (approved):
LGTM so far!

src/widgets/attachments/attachments-overview.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+  ] = useCurrentPatient();
+
+  useEffect(() => {
+    if (patientUuid) {
+      const abortController = new AbortController();
+      getAttachments(patientUuid, true, abortController).then(
+        (response: any) => {
+          const listItems = response.data.results.map(attachment => ({
+            src: `/openmrs/ws/rest/v1/attachment/${attachment.uuid}/bytes`,
+            thumbnail: `/openmrs/ws/rest/v1/attachment/${attachment.uuid}/bytes`,
```
ivange94, line 28: Right now I'm using the source of the original image as the `src` value and `thumbnail` value. 
When [ATT-38 AttachmentBytesResource_* to take the (complex obs) view as parameter](https://issues.openmrs.org/browse/ATT-38) is complete, I should be able to access the thumbnail bytes.

src/widgets/attachments/attachments-overview.component.tsx
```diff
@@ -0,0 +1,122 @@
...
+            src: `/openmrs/ws/rest/v1/attachment/${response.data.uuid}/bytes`,
+            thumbnail: `/openmrs/ws/rest/v1/attachment/${response.data.uuid}/bytes`,
+            thumbnailWidth: 320,
+            thumbnailHeight: 212,
+            caption: response.data.comment
+          };
+          attachments_tmp.push(new_attachment);
+        });
+      }
+      setAttachments(attachments_tmp);
```
FlorianRappl: Is this working?

Reason I am asking: Populating the `attachments_tmp` is async (i.e., may / will happen in the next cycle), while `setAttachments` happens in the current cycle. As such - even though it may seem to work locally - it is not reliable.

Rather we should use

```js
const result = Promise.all(files.map(file => createAttachment(
          patientUuid,
          file,
          file.name,
          abortController
        ).then(response => {
          const new_attachment = {
            id: `${response.data.uuid}`,
            src: `/openmrs/ws/rest/v1/attachment/${response.data.uuid}/bytes`,
            thumbnail: `/openmrs/ws/rest/v1/attachment/${response.data.uuid}/bytes`,
            thumbnailWidth: 320,
            thumbnailHeight: 212,
            caption: response.data.comment
          };
          attachments_tmp.push(new_attachment);
        })));
result.then(() => setAttachments(attachments_tmp));
```

src/widgets/attachments/attachments-overview.component.tsx
```diff
@@ -0,0 +1,122 @@
...
+    }
+  }
+
+  function handleCurrentImageChange(index) {
+    setCurrentImage(index);
+  }
+
+  function handleDelete() {
+    if (window.confirm("Are you sure you want to delete attachment?")) {
```
FlorianRappl, line 73: I guess this is right now only for dev convenience, right?

Do we have a UX strategy how to deal with such dialogs or does that come back to the need for a modal manager @brandones ?

FlorianRappl (commented):
I think we are on a good way here.

Great work so far @ivange94 !

src/widgets/attachments/attachments-overview.component.tsx
ivange94: Thanks I will fix that. Actually I noticed a bug. I had to refresh the page to see the newly uploaded attachment.

src/widgets/attachments/attachments-overview.component.tsx
ivange94: This was my first instinct but I think we should have a standard for doing these kind of things.

FlorianRappl: I fully agree. Not sure if its on our radar design-wise, but I will also consider it for the upcoming modal dialog manager RFC.

brandones: Sure, having a technical strategy for this would be good.

I would also like us to do some thinking (with Cairan's help, if possible) about "UX strategy" in a more broad sense -- when are dialog popups really necessary, what must they contain to be worthwhile, how can they be avoided, etc.? For example, once we have a proper modal, we should display the thumbnail of the attachment being deleted in the dialog box.

FlorianRappl (approved):
LGTM!

Great work and thanks for the presentation yesterday :rocket:

jonathandick:
Is this ready to be merged? If so, @ivange94 could you fix the failing travis check?

ivange94:
@jonathandick yes this is ready for merge. I've fixed the linting errors. Will I need to squash the commits too?

brandones:
Yes, but you can squash from right here in the UI

screenshot--2020-05-11--101228

Feel free to edit the commit message details if you think it could be more helpful/informative about what you actually did.

Nice work @ivange94 !

src/widgets/attachments/attachments-overview.component.tsx
```diff
@@ -0,0 +1,119 @@
...
+  ] = useCurrentPatient();
+
+  useEffect(() => {
+    if (patientUuid) {
+      const abortController = new AbortController();
+      getAttachments(patientUuid, true, abortController).then(
+        (response: any) => {
+          const listItems = response.data.results.map(attachment => ({
+            id: `${attachment.uuid}`,
+            src: `/openmrs/ws/rest/v1/attachment/${attachment.uuid}/bytes`,
```
dkayiwa, line 29: @ivange94 @brandones @FlorianRappl will this work when the context path is not `openmrs`?

FlorianRappl: No it will not work and we should not build urls like this.

PR #41, opened 2020-04-16
MF-168: Fix unique key prop error for vitals resource

denniskigen:
Screenshot 2020-04-16 at 13 03 54

All the vitals had an id `undefined` as the property with the encounter id in FHIR is called `encounter`, not `context`.

Screenshot 2020-04-16 at 13 03 30

Steps to recreate:

- Load up patient search and search for 'John Wilson' or any other patient with recorded vitals.
- The error should pop up in the console.


donaldkibet (approved):
lgtm

nkimaina (approved):

PR #40, opened 2020-04-15
Fix servicetype initial state and location value.

drizzentic:

PR #42, opened 2020-04-16
MF-169: Fix failing CI builds

denniskigen:
Test suite runs are failing in Travis with this error:

Cannot find module `@babel/compat-data/corejs3-shipped-proposals`

This is due to a breaking change in node 13.13 which breaks with Babel <7.9.0 https://github.com/babel/babel/issues/11427#issuecomment-614584591

This PR updates the following packages to the latest versions to fix the breaking change:

`@babel/core`
`@babel/plugin-proposal-optional-chaining`
`@babel/preset-env`
`@babel/preset-react`
`@babel/preset-typescript`
`babel-loader`


drizzentic:
@denniskigen Try adding `npm update` in the Travis config and it should fix it. I had the same issue yesterday on https://github.com/openmrs/openmrs-esm-patient-chart-widgets/pull/40

FlorianRappl:
> @denniskigen Try adding npm update in the Travis config and it should fix it. I had the same issue yesterday on #40

Not sure if that's the best suggestion. Shouldn't Travis see the same thing as we do locally? This way Travis will always get the latest versions as per package.json rules. Usually, the package-lock.json should be respected.

Maybe run `npm update` locally and then pushing the changed package-lock.json?

denniskigen:
> > @denniskigen Try adding npm update in the Travis config and it should fix it. I had the same issue yesterday on #40
> 
> Not sure if that's the best suggestion. Shouldn't Travis see the same thing as we do locally? This way Travis will always get the latest versions as per package.json rules. Usually, the package-lock.json should be respected.
> 
> Maybe run `npm update` locally and then pushing the changed package-lock.json?

Done and the build passed @FlorianRappl.

PR #43, opened 2020-04-16
MF-170 : Switched Height and Weight Resource from REST API to FHIR 2

donaldkibet:


drizzentic (commented):
LGTM

kagai (commented):
@donaldkibet good work so far.  Is the data from FHIR the same as that from openmrs endpoint ,if not i think you will need to change how the data is displayed on the components

src/widgets/heightandweight/heightandweight.resource.ts
```diff
@@ -5,72 +5,50 @@ import { formatDate, calculateBMI } from "./heightandweight-helper";
 const HEIGHT_CONCEPT = "5090AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
```
nkimaina, line 5: @drizzentic In another PR/ticket, we need to move this to a config file.

nkimaina (approved):

PR #44, opened 2020-04-21
MF-182: Empty states for widgets

denniskigen:
This PR introduces an empty state for widgets to render when they do not have data to show yet.

Presently, most widgets will render an incomplete view. For example, in the case of the Notes widget, table headers will be displayed without the corresponding table data.

Screenshot 2020-04-26 at 22 15 45

This could be improved upon by rendering an empty state view that communicates that there are no recorded notes for the patient as well as an affordance that would allow the user to record new notes.

Screenshot 2020-04-26 at 22 19 23


FlorianRappl (commented):
There are quite some changes (38 files). Note sure if all is required - I would still love to review.

Can you fix the merge conflicts please? Thanks!

denniskigen:
> There are quite some changes (38 files). Note sure if all is required - I would still love to review.
> 
> Can you fix the merge conflicts please? Thanks!

Fixed

src/widgets/vitals/vitals-overview.component.tsx
```diff
@@ -54,76 +55,86 @@ export default function VitalsOverview(props: VitalsOverviewProps) {
...
+                <td>BP</td>
+                <td>Rate</td>
+                <td>Oxygen</td>
+                <td colSpan={2}>Temp</td>
+              </tr>
+            </thead>
+            <tbody>
+              {currentVitals.map((vital, index) => {
+                return (
+                  <React.Fragment key={vital.id}>
```
FlorianRappl: We only return the row, right? I would remove the fragment and only use `<tr key={vital.id}>`.

src/widgets/heightandweight/heightandweight-record.component.tsx
```diff
@@ -76,7 +76,7 @@ export default function HeightAndWeightRecord(
                   <tr>
                     <td>Weight</td>
                     <td>
-                      {dimensions.weight} <span>Kg</span>
+                      {dimensions.weight} <span>kg</span>
```
FlorianRappl, line 79: Can we store these constant labels somewhere (e.g., in a constants.ts file)?

I assume that the API always gives us `kg`, right?

FlorianRappl: Above we use `dimensions?.obsData` here `dimensions.weight`. Can `dimensions` be `undefined` or not?

src/ui-components/empty-state/empty-state.component.tsx
```diff
@@ -0,0 +1,28 @@
...
+    </SummaryCard>
+  );
+}
+
+type EmptyStateProps = {
+  name: string;
+  displayText: string;
+  styles?: React.CSSProperties;
+  addComponent?: string | any;
+  showComponent?: Function;
```
FlorianRappl: Can we type this function? What is it (e.g., `() => void`)?

src/ui-components/empty-state/empty-state.component.tsx
```diff
@@ -0,0 +1,28 @@
...
+      </div>
+    </SummaryCard>
+  );
+}
+
+type EmptyStateProps = {
+  name: string;
+  displayText: string;
+  styles?: React.CSSProperties;
+  addComponent?: string | any;
```
FlorianRappl: Well, `string | any` is `any`. Can we be more specific?

src/widgets/vitals/vitals-overview.component.tsx
denniskigen: Yeah, good shout. Fixed.

src/widgets/heightandweight/heightandweight-record.component.tsx
denniskigen: It can be `undefined` if the record with the specified id isn't found. I've rectified this.

src/ui-components/empty-state/empty-state.component.tsx
denniskigen: Yes, `() => void` this would do.

src/ui-components/empty-state/empty-state.component.tsx
denniskigen: Yes we can. I think it can be typed to:

```
import { DataCaptureComponentProps } from "../../widgets/shared-utils";
import { match } from "react-router-dom";  

type EmptyStateProps = {
  name: string;
  displayText: string;
  styles?: React.CSSProperties;
  addComponent?: React.FC<TProps | DataCaptureComponentProps>;
  showComponent?: () => void;
};

type TProps = {
  basePath?: string;
  match?: match;
}
```

src/widgets/heightandweight/heightandweight-record.component.tsx
denniskigen: > Can we store these constant labels somewhere (e.g., in a constants.ts file)?
> 
> I assume that the API always gives us `kg`, right?

I think so. I'm not sure where the appropriate location for this would be. Perhaps @nkimaina can chime in on this?

src/ui-components/empty-state/empty-state.component.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+type EmptyStateProps = {
+  name: string;
+  displayText: string;
+  styles?: React.CSSProperties;
+  addComponent?: React.FC<TProps | DataCaptureComponentProps>;
+  showComponent?: () => void;
+};
+
+type TProps = {
```
FlorianRappl: When I see `TProps` I always assume its a generic type. Can we give this a more accurate name, e.g., `RouteBaseComponentProps`?

src/widgets/allergies/allergy-form.component.tsx
```diff
@@ -666,12 +666,12 @@ export default function AllergyForm(props: AllergyFormProps) {
   }
 
   React.useEffect(() => {
-    if (props.match.params["allergyUuid"]) {
+    if (props.match?.params?.["allergyUuid"]) {
```
FlorianRappl: Can be simplified:

```ts
setViewForm(!!props.match?.params?.["allergyUuid"]);
```

src/widgets/appointments/appointment-record.component.tsx
```diff
@@ -127,18 +127,18 @@ export default function AppointmentRecord(props: AppointmentRecordProps) {
               <tbody>
                 <tr>
                   <td>
-                    {dayjs(patientAppointment.startDateTime).format(
+                    {dayjs(patientAppointment?.startDateTime).format(
                       "DD-MMM-YYYY"
                     )}
                   </td>
                   <td>
                     {patientAppointment.service.creatorName
```
FlorianRappl: I guess we miss the conditional member operator here `?.`. Actually I think the three lines could be replaced by

```ts
patientAppointment?.service?.creatorName || "\u2014"
```

src/widgets/appointments/appointments.component.tsx
```diff
@@ -3,7 +3,7 @@ import { Switch, Route, useRouteMatch } from "react-router-dom";
 import AppointmentsDetailedSummary from "./appointments-detailed-summary.component";
 import AppointmentRecord from "./appointment-record.component";
 
-export default function Appointements() {
+export default function Appointments() {
```
FlorianRappl, line 6: Great catch! :eyes:!

src/widgets/heightandweight/heightandweight-overview.component.tsx
```diff
@@ -37,62 +40,91 @@ export default function HeightAndWeightOverview(
...
+                    <th className={styles.tableHeader}>Height</th>
+                    <th className={styles.tableHeader}>BMI</th>
+                    <th></th>
+                  </tr>
+                </thead>
+                <tbody>
+                  {dimensions
+                    .slice(0, showMore ? 6 : 3)
+                    .map((dimension, index) => (
+                      <tr key={dimension.id} className={styles.tableRow}>
```
FlorianRappl, line 70: Hehe, I'm always looking for missing `key` props - but I never find them! This is what I call careful coding. Great stuff! 🚀

src/widgets/heightandweight/heightandweight-record.component.tsx
FlorianRappl: Ping @nkimaina - thanks!

src/widgets/heightandweight/heightandweight-util.tsx
```diff
@@ -0,0 +1,14 @@
+import { newWorkspaceItem } from "@openmrs/esm-api";
+
+export function openHeightAndWeightTab(addComponent, componentName): void {
```
FlorianRappl: Hehe - the `void` is known anyway (we don't return anything), but `addComponent` and `componentName` still need a type.

src/widgets/medications/medications-overview.component.tsx
```diff
@@ -198,6 +171,128 @@ export default function MedicationsOverview(props: MedicationsOverviewProps) {
...
+                    </span>{" "}
+                    &mdash;{" "}
+                    <span>{(medication?.route?.display).toLowerCase()}</span>{" "}
+                    <span
+                      style={{
+                        fontWeight: 400,
+                        color: "var(--omrs-color-ink-high-contrast)"
+                      }}
+                    >
+                      &mdash; {medication.frequency.display}
```
FlorianRappl: Missing conditional operator?

src/widgets/medications/medications-overview.component.tsx
```diff
@@ -198,6 +171,128 @@ export default function MedicationsOverview(props: MedicationsOverviewProps) {
...
+                        medication?.dose
+                      ).toLowerCase()}
+                    </span>{" "}
+                    &mdash;
+                    <span>
+                      {" "}
+                      {(medication?.doseUnits?.display).toLowerCase()}
+                    </span>{" "}
+                    &mdash;{" "}
+                    <span>{(medication?.route?.display).toLowerCase()}</span>{" "}
```
FlorianRappl: Potential undefined `toLowerCase` -> that call needs to be in the chain or used with `?.`.

src/widgets/medications/medications-overview.component.tsx
```diff
@@ -198,6 +171,128 @@ export default function MedicationsOverview(props: MedicationsOverviewProps) {
...
+                    >
+                      {getDosage(
+                        medication?.drug?.strength,
+                        medication?.dose
+                      ).toLowerCase()}
+                    </span>{" "}
+                    &mdash;
+                    <span>
+                      {" "}
+                      {(medication?.doseUnits?.display).toLowerCase()}
```
FlorianRappl: I would remove these parentheses.

Note: Its a difference if you write

```ts
(medication?.doseUnits?.display).toLowerCase()
medication?.doseUnits?.display.toLowerCase()
```

the former always calls `toLowerCase` (and thus runs it potentially on undefined) while the latter has it in the chain, i..e, it only runs it if the conditionals are fulfilled.

src/widgets/medications/medications-overview.component.tsx
```diff
@@ -198,6 +171,128 @@ export default function MedicationsOverview(props: MedicationsOverviewProps) {
...
+                    >
+                      &mdash; {medication.frequency.display}
+                    </span>
+                  </td>
+                  <td>
+                    <MedicationButton
+                      component={MedicationOrderBasket}
+                      name={"Medication Order Basket"}
+                      label={"Revise"}
+                      orderUuid={medication.uuid}
```
FlorianRappl: From here on there are no conditional members - maybe `medication` is always *not* `undefined`?

FlorianRappl (changes requested):
This is a long one. In general it looks good to me - there are just some improvements for the conditional member operator etc. that would make sense.

Great work :rocket!

src/widgets/medications/medications-overview.component.tsx
denniskigen: Sure. Good spot.

src/widgets/medications/medications-overview.component.tsx
denniskigen: Excellent

src/ui-components/empty-state/empty-state.component.tsx
denniskigen: Sure. That's a great variable name. Also is it `RouteBase` or `RouteBased`?

src/widgets/allergies/allergy-form.component.tsx
denniskigen: Concise but I'd argue harder to read.

src/widgets/appointments/appointment-record.component.tsx
denniskigen: Yeah. Elegant solution.

src/widgets/heightandweight/heightandweight-util.tsx
denniskigen: Surprised - I thought I got rid of this file in a previous commit. Must have rebased improperly. `openHeightAndWeightTab` should be replaced entirely by the reusable (and correctly typed) `openWorkspaceTab` function. I'll fix this.

src/widgets/medications/medications-overview.component.tsx
denniskigen: Yes. Good spot.

src/widgets/medications/medications-overview.component.tsx
denniskigen: Fixed as suggested below 💯

jonathandick:
@denniskigen what's the status of this PR? @FlorianRappl is this approved? Can we merge?

jonathandick:
Ok. I'm going to merge this as the travis issue is fixed.

PR #45, opened 2020-04-22
Start visit

nkimaina:


src/utils/use-session-user.tsx
```diff
@@ -0,0 +1,17 @@
...
+export default function useSessionUser() {
+  const [sessionUser, setSessionUser] = useState<any>(null);
+  useEffect(() => {
+    let currentUserSub: Subscription;
+    if (sessionUser === null) {
+      openmrsObservableFetch("/ws/rest/v1/session").subscribe(user => {
+        setSessionUser(user.data);
+      });
+    }
+    return () => currentUserSub.unsubscribe();
```
FlorianRappl: This will fail - currentUserSub may (i.e., will) be undefined

src/utils/use-session-user.tsx
```diff
@@ -0,0 +1,17 @@
+import { useState, useEffect } from "react";
+import { Subscription } from "rxjs";
+import { openmrsObservableFetch } from "@openmrs/esm-api";
+
+export default function useSessionUser() {
+  const [sessionUser, setSessionUser] = useState<any>(null);
+  useEffect(() => {
+    let currentUserSub: Subscription;
+    if (sessionUser === null) {
+      openmrsObservableFetch("/ws/rest/v1/session").subscribe(user => {
```
FlorianRappl: I think you want to set `currentUserSub` here, right?

src/widgets/location/location-select.component.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export default function LocationSelect(props: LocationSelectProps) {
+  const [locations] = useLocations();
+  const onLocationsChanged = event => {
+    props.onLocationChanged(
+      locations.find(loc => loc.uuid == event.target.value)
+    );
+  };
+  return (
+    <>
```
FlorianRappl: Is the fragment really necessary in this case? There is just one child element.

src/widgets/location/location-select.component.tsx
```diff
@@ -0,0 +1,45 @@
...
+      <select
+        name="visitLocation"
+        id={props.id || "visitLocation"}
+        className="omrs-type-body-regular"
+        style={{ height: "40px" }}
+        value={props.currentLocationUuid}
+        onChange={onLocationsChanged}
+      >
+        <option value={null} className="omrs-padding-8">
+          {""}
```
FlorianRappl: I think this is the same as `<option value="" className="omrs-padding-8" />`

src/widgets/location/use-locations.tsx
```diff
@@ -0,0 +1,20 @@
+import { useState, useEffect } from "react";
+import { Subscription } from "rxjs";
+import { Location, getLocations } from "./location.resource";
+
+export default function useLocations() {
+  const [locations, setLocations] = useState<Array<Location>>([]);
+  useEffect(() => {
+    let locationSub: Subscription;
```
FlorianRappl: Why not just `const locationSub = ...`? I think the typing would also benefit (no need for the explicit `Subscription`)

src/widgets/visit/use-started-visit.tsx
```diff
@@ -0,0 +1,60 @@
+import React, { useState, useEffect } from "react";
+import { Subscription } from "rxjs";
+import dayjs, { Dayjs } from "dayjs";
+import { getCurrentPatientUuid } from "@openmrs/esm-api";
+
+import { getVisitsForPatient } from "./visit.resource";
+
+export default function useStartedVisit(props: StartedVisitProps) {
+  const [startedVisit, setStartedVisit] = useState(undefined);
```
FlorianRappl: `useState<?>(undefined)` - what is it?

src/widgets/visit/use-visit-types.tsx
```diff
@@ -0,0 +1,20 @@
+import { useState, useEffect } from "react";
+import { Subscription } from "rxjs";
+import { VisitType, getVisitTypes } from "./visit-type.resource";
+
+export default function useVisitTypes() {
+  const [visitTypes, setVisitTypes] = useState<Array<VisitType>>([]);
+  useEffect(() => {
+    let visitTypesSub: Subscription;
+    visitTypesSub = getVisitTypes().subscribe(
```
FlorianRappl: See above - just assign in initialization.

src/widgets/visit/visit-type-select.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+    <>
+      <select
+        name="visitType"
+        id={props.id || "visitType"}
+        className="omrs-type-body-regular"
+        style={{ height: "40px" }}
+        onChange={onVisitTypesChanged}
+      >
+        <option value={null} className="omrs-padding-8">
+          {""}
```
FlorianRappl: Same as above.

FlorianRappl (changes requested):
I think we are on a great track here. Thanks for the work!

nkimaina:
@FlorianRappl Sorry that these changes are a lot to review, we are transferring this functionality from an internal repo. Also, we've not had a design session to agree on how visits should function, but these are standard features that will most likely be needed when we have the design.

donaldkibet (approved):
lgtm

PR #47, opened 2020-04-24
MF-184: Make newWorkspaceItem reusable across widgets

denniskigen:
Presently, most widgets are redeclaring the logic for launching new workspaces. It would be worthwhile to DRY it up by making that bit of logic reusable.


drizzentic (commented):
LGTM

PR #48, opened 2020-04-24
MF-184-allergies: Make newWorkspaceItem reusable across allergies widget

denniskigen:
Make newWorkspaceItem reusable across the components that make up the allergies widget.

### DIFF
--- a/src/widgets/shared-utils.tsx
+++ b/src/widgets/shared-utils.tsx
@@ -1,0 +1,14 @@
+import { newWorkspaceItem } from "@openmrs/esm-api";
+
+export const openWorkspaceTab = (componentToAdd, componentName, params?) => {
+  newWorkspaceItem({
+    component: componentToAdd,
+    name: componentName,
+    props: {
+      match: { params: params ? params : {} }
+    },
+    inProgress: false,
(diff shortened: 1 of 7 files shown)


src/widgets/shared-utils.tsx
```diff
@@ -0,0 +1,14 @@
+import { newWorkspaceItem } from "@openmrs/esm-api";
```
drizzentic, line 1: Duplication of this https://github.com/openmrs/openmrs-esm-patient-chart-widgets/pull/47/files

drizzentic (changes requested):

src/widgets/shared-utils.tsx
denniskigen: Yeah, about this, I was trying out a different approach with this PR. Often when making large-scale refactors such as this one, you'll find that PR sizes get unwieldy with huge numbers of files changed. I was trying out this approach Joel suggested for keeping merge requests [small and manageable](https://docs.google.com/document/d/1_jCuT3vRSW8WKJbT9manso_VM0JoUIb6HpFmBi_yDMs/edit). Sorry for the repetition.

PR #50, opened 2020-04-24
MF-184-dimensions: Make newWorkspaceItem reusable across height & weight widget

denniskigen:
Make newWorkspaceItem reusable across the components that make up the height and weight widget.

### DIFF
--- a/src/widgets/shared-utils.tsx
+++ b/src/widgets/shared-utils.tsx
@@ -1,0 +1,14 @@
+import { newWorkspaceItem } from "@openmrs/esm-api";
+
+export const openWorkspaceTab = (componentToAdd, componentName, params?) => {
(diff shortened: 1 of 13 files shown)


src/widgets/shared-utils.tsx
```diff
@@ -0,0 +1,14 @@
+import { newWorkspaceItem } from "@openmrs/esm-api";
+
+export const openWorkspaceTab = (componentToAdd, componentName, params?) => {
```
FlorianRappl, line 3: This is the same as beforehand, right? Maybe just fix the other branch and update this PR. Then everything should be fine!

denniskigen: Yeah, let me close this one since the changes have already been merged in another PR.

PR #51, opened 2020-04-24
MF-184-medications: Make newWorkspaceItem reusable across medications widget 

denniskigen:
Make newWorkspaceItem reusable across the components that make up the medications widget.

### DIFF
--- a/src/widgets/shared-utils.tsx
+++ b/src/widgets/shared-utils.tsx
@@ -1,0 +1,14 @@
+import { newWorkspaceItem } from "@openmrs/esm-api";
+
+export const openWorkspaceTab = (componentToAdd, componentName, params?) => {
+  newWorkspaceItem({
+    component: componentToAdd,
+    name: componentName,
+    props: {
+      match: { params: params ? params : {} }
+    },
+    inProgress: false,
+    validations: (workspaceTabs: Array<{ component: React.FC }>) =>
+      workspaceTabs.findIndex(tab => tab.component === componentToAdd)
+  });
+};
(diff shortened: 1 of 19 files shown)


src/widgets/shared-utils.tsx
```diff
@@ -0,0 +1,14 @@
+import { newWorkspaceItem } from "@openmrs/esm-api";
+
+export const openWorkspaceTab = (componentToAdd, componentName, params?) => {
```
FlorianRappl, line 3: Can you type these params?

Maybe use

```ts
export function openWorkspaceTab(componentToAdd: React.FC, componentName: string, params?: any) {
}
```

- What is `params`
- Is this a type without props? If not, what are these props?
```

FlorianRappl (changes requested):
Looks good - just needs more type decl.

src/widgets/shared-utils.tsx
denniskigen: Sure, params is an Object. It is used for passing around stuff like condition/order uuids, or condition/drug/etc names to the form when editing individual records. I'll close this PR since the changes it contained have already been merged and then open a different one addressing the typing of `params` and `componentToAdd`.

denniskigen: > Can you type these params?
> 
> Maybe use
> 
> ```ts
> export function openWorkspaceTab(componentToAdd: React.FC, componentName: string, params?: any) {
> }
> ```
> 
> * What is `params`
> * Is this a type without props? If not, what are these props?

Hey @FlorianRappl, I've addressed these concerns in #58 .

PR #57, opened 2020-04-27
MF-184-conditions: Make newWorkspaceItem reusable in the conditions component

denniskigen:
Make`newWorkspaceItem` reusable in the components that make up the Conditions widget.


src/widgets/conditions/condition-record.component.tsx
```diff
@@ -32,14 +32,9 @@ export default function ConditionRecord(props: ConditionRecordProps) {
...
-            openEditConditionsWorkspaceTab(
-              ConditionsForm,
-              "Edit Conditions",
-              patientCondition?.id,
-              patientCondition?.code?.text,
-              patientCondition?.clinicalStatus,
-              patientCondition?.onsetDateTime
-            );
+            openWorkspaceTab(ConditionsForm, "Edit Conditions", {
+              conditionUuid: patientCondition?.id
```
FlorianRappl: Is that equivalent? Why are the props like the clinical status etc. not required any more?

(Just for my understanding - technically this seems to be alright)

FlorianRappl (commented):
LGTM so far!

src/widgets/conditions/condition-record.component.tsx
denniskigen: I'm making a request in `ConditionsForm` component to fetch a condition using the provided uuid. I could as well pass the data required from the props in `ConditionRecord`. Which approach would you prefer?

denniskigen: ping @FlorianRappl

FlorianRappl: Sorry @denniskigen - I lost my track here.

Back to your question: If the data is already available we should just pass it on (instead of making another request). Otherwise, having the request only in `ConditionsForm` makes sense to me. So if its the latter - leave it, if we do have all data then pass it on.

denniskigen: No worries. We have all the data. I'll refactor it so that it passes on the data then. Thanks.

FlorianRappl: Great thanks! :+1:

jonathandick:
@denniskigen this has a failing travis check. Could you take a look?

denniskigen:
> @denniskigen this has a failing travis check. Could you take a look?

That failing test is fixed by #64. That needs to be merged first.

denniskigen:
Mergeable now @jonathandick

PR #58, opened 2020-04-27
MF-184-wrapup: Properly type shared utils

denniskigen:
Final bits of work on https://issues.openmrs.org/browse/MF-184.

Properly type `openWorkspaceTab` and `DataCaptureComponentProps`. Also, `DataCaptureComponentProps` is now reusable.

### DIFF
--- a/src/widgets/shared-utils.tsx
+++ b/src/widgets/shared-utils.tsx
@@ -1,10 +1,9 @@
 import { newWorkspaceItem } from "@openmrs/esm-api";
 
-export function openWorkspaceTab<TProps, TParams = any>(
-  componentToAdd: React.FC<TProps | DataCaptureComponentProps>,
-  componentName: string,
-  params?: TParams
-) {
+export function openWorkspaceTab<
+  TProps = DataCaptureComponentProps,
+  TParams = any
+>(componentToAdd: React.FC<TProps>, componentName: string, params?: TParams) {
   newWorkspaceItem({
     component: componentToAdd,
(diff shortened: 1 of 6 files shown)


src/widgets/shared-utils.tsx
```diff
@@ -1,6 +1,10 @@
 import { newWorkspaceItem } from "@openmrs/esm-api";
 
-export const openWorkspaceTab = (componentToAdd, componentName, params?) => {
+export function openWorkspaceTab<TProps, TParams = any>(
```
FlorianRappl: Maybe use `TProps = DataCaptureComponentProps` (i.e., use `DataCaptureComponentProps` as default).

Below you state that the functional component has props that are either derived into `TProps` or are `DataCaptureComponentProps`; which means it will always just be `TProps` and in some cases `TProps = DataCaptureComponentProps`. The above takes that away with a default value.

FlorianRappl (commented):
I think its a great change. Would just change the signature as noted, but then we can merge.

src/widgets/shared-utils.tsx
denniskigen: That's very clear, thanks. I've changed it to be so.

FlorianRappl (approved):
LGTM! Thanks for awesome work :rocket:!

denniskigen:
> LGTM! Thanks for awesome work :rocket:!

Thanks for the help with types and thanks for reviewing my code!

PR #60, opened 2020-04-29
MF-186

kagai:
We have different concept UUIDs for the different organizations involved in the development of MF. This config is the first draft of how we can manage these, ideas and improvements are welcome.

### DIFF
--- a/src/concept-config.json
+++ b/src/concept-config.json
@@ -1,0 +1,27 @@
+{
+  "AMPATH": {
+    "concepts": [
+      {
+        "uuid": "",
+        "name": "DURATION UNITS CONCEPT"
+      },
+      {
+        "uuid": "",
+        "name": "CARE SETTING"
+      }
+    ]
+  },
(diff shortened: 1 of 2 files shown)


nkimaina:
Generally, see how to do configs using ems-config module. See example for login: https://github.com/openmrs/openmrs-esm-login/blob/master/src/root.component.tsx#L10 and how we provide our own config at AMPATH for the same: https://github.com/AMPATH/ampath-esm-root-config/blob/master/src/configs/login.config.json and load it: https://github.com/AMPATH/ampath-esm-root-config/blob/master/src/root-config.ts#L11

src/concept-config.json
```diff
@@ -0,0 +1,27 @@
+{
+  "AMPATH": {
```
nkimaina, line 2: We don't need to define organization specific like this, if we use the esm-config to define the config files, as it provides a way for each org/implementation to provide their own config.

PR #64, opened 2020-05-06
MF-190: Fix off-by-one date bug that was causing test to fail

denniskigen:
Screenshot 2020-05-06 at 15 30 30

Thanks @FlorianRappl for helping with this.


FlorianRappl (approved):
LGTM!

PR #65, opened 2020-05-11
Chart base path hook

nkimaina:


nkimaina:
This hook will be important for majority of components that will be built with openmrs repos and third party modules by various implementations.

src/utils/use-chart-base.ts
```diff
@@ -0,0 +1,9 @@
+import { useRouteMatch } from "react-router-dom";
+export default function useChartBasePath() {
+  const match = useRouteMatch();
+  const chartBasePath = `${match.url.substr(
+    0,
+    match.url.search("/chart/")
+  )}/chart`;
+  return [chartBasePath];
```
FlorianRappl: Why do we package this in an array?

FlorianRappl (commented):
In general looks good, but I am not sure why we pack (and then need to unpack) this in an array.

FlorianRappl:
> This hook will be important for majority of components that will be built with openmrs repos and third party modules by various implementations.

I agree, but I think that such a hook should not be necessary in general. Right now this hook assumes that a certain part of the URL (`/chart`) is always there and needed. While this, of course, can be changed per repo, I wonder if we really should rely on this.

nkimaina:
> I agree, but I think that such a hook should not be necessary in general. Right now this hook assumes that a certain part of the URL (/chart) is always there and needed. While this, of course, can be changed per repo, I wonder if we really should rely on this

Agreed. It occurred to me later on that I can't actually use that hook in a 3rd party module because of issues with react context. I have removed the export part, so that it's used within this module only. Could we just rely on these changes as a refactor to avoid duplication within this repo, or would that be premature abstraction?

FlorianRappl:
> It occurred to me later on that I can't actually use that hook in a 3rd party module because of issues with react context.

Well the specific hook you can, however, how would the constant (`/chart/`) be determined? If one would need to pass that in always then it would not work. The context will always be based on the closest context - so this is no problem as long as `react-router` is used.

> I have removed the export part, so that it's used within this module only.

This comment I do not understand. You'll need to export it from a module in order to consume it in another. All these modules should be "internal" anyway as in a MFE we don't export anything.

> Could we just rely on these changes as a refactor to avoid duplication within this repo, or would that be premature abstraction?

I think my comment was misunderstood. What I was referring to is that we should not need such a hook or any special URL treatment. If our solution is proper, then we don't need any additional abstraction. Just using `react-router` as it should be used is the way to do it. Anything else is prone to error and requires custom knowledge. It should not be our intention to re-invent the wheel. We should use established techniques as much as possible. Unfortunately, our current architecture requires many custom things. That's what my comment was about.

FlorianRappl:
I think in general we need to talk about  `*-widgets`.

If we keep on adding application logic (this includes routing) in the widgets then the potential re-use is close to zero. Why do we have two repos (`openmrs-esm-patient-chart-widgets` and `openmrs-esm-patient-chart`) when we can have one? If we want to allow re-use, then we urgently need to refactor the included components to be more general.

As a matter of fact, all these components should be design+content only. No API logic. If we call the API or perform routing, then we should rely on some props or a context.

cc @brandones @jonathandick

FlorianRappl (approved):

brandones:
Yeah @FlorianRappl , it's a very interesting point -- wanna cross-post your comment to Slack and then we'll continue the conversation there?

PR #66, opened 2020-05-12
MF-201: Fixes to the Add New Program logic

denniskigen:
https://issues.openmrs.org/browse/MF-201

Fixes:

1. User being unable to select a different location than the default session location.
2. Sorts enrolled programs by enrollment date so that the newest enrollment is displayed at the top in the detailed summary view.

### DIFF
--- a/src/widgets/programs/programs-detailed-summary.component.tsx
+++ b/src/widgets/programs/programs-detailed-summary.component.tsx
@@ -55,46 +55,41 @@ export default function ProgramsDetailedSummary(
           </thead>
           <tbody>
             {enrolledPrograms &&
-              enrolledPrograms
-                .slice()
-                .sort((a, b) => (b.dateEnrolled > a.dateEnrolled ? 1 : -1))
-                .map(program => {
-                  return (
-                    <React.Fragment key={program.uuid}>
-                      <tr
-                        className={`${
-                          program.dateCompleted
-                            ? `${styles.inactive}`
-                            : `${styles.active}`
-                        }`}
-                      >
-                        <td className="omrs-medium">{program.display}</td>
-                        <td>
-                          {dayjs(program.dateEnrolled).format("MMM-YYYY")}
-                        </td>
-                        <td>
-                          {program.dateCompleted
-                            ? `Completed on ${dayjs(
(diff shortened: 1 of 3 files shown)


src/widgets/programs/programs-detailed-summary.component.tsx
```diff
@@ -56,41 +55,46 @@ export default function ProgramsDetailedSummary(
...
-                          </Link>
-                        }
-                      </td>
-                    </tr>
-                  </React.Fragment>
-                );
-              })}
+              enrolledPrograms
+                .slice()
+                .sort((a, b) => (b.dateEnrolled > a.dateEnrolled ? 1 : -1))
```
FlorianRappl: Should we perform all these changes already when setting the `enrolledPrograms`? Otherwise, we need to recompute every time we render. Alternatively, use `useMemo` to get a memoized version of it, with the memoization key being set to `enrolledPrograms`.

FlorianRappl (commented):
LGTM so far!

src/widgets/programs/programs-detailed-summary.component.tsx
denniskigen: Probably not. Perhaps it would be more appropriate to move this concern to `programs.resource` so that `fetchEnrolledPrograms` would always return a sorted collection:

```
export function fetchEnrolledPrograms(
  patientID: string
): Observable<PatientProgram[]> {
  return openmrsObservableFetch(
    `/ws/rest/v1/programenrollment?patient=${patientID}`
  ).pipe(
    map(
      ({ data }) =>
        data["results"].sort((a, b) =>
          b.dateEnrolled > a.dateEnrolled ? 1 : -1
        ),
      take(3)
    )
  );
}
```

FlorianRappl: Sounds good to me!

FlorianRappl (approved):

src/widgets/programs/programs-detailed-summary.component.tsx
denniskigen: Thanks @FlorianRappl!

PR #67, opened 2020-05-12
MF-161: fixed revised/expired  orders not appearing in past meds

donaldkibet:
This PR addresses showing expired and revised medication, but doesn't show discontinued orders because of the API has a bug not displaying discontinued orders
Screenshot from 2020-05-12 15-24-35


FlorianRappl (approved):
LGTM!

PR #68, opened 2020-05-12
MF-201: Fix issue with future dates when adding programs

denniskigen:
https://issues.openmrs.org/browse/MF-201

Fixes issue where the user could be allowed to select a date in the future as the enrollment date in the Date Enrolled input field. 

Rather than fleshing out fully-fledged form validation logic, this PR relies on the [Constraint Validation API](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/Constraint_validation) and tries to build out some basic [date input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#Setting_maximum_and_minimum_dates) validation.

Screenshot 2020-05-12 at 17 56 40

Translations working:

Screenshot 2020-05-15 at 20 29 51

### DIFF
--- a/src/widgets/programs/programs-form.component.tsx
+++ b/src/widgets/programs/programs-form.component.tsx
@@ -239,6 +239,7 @@ export default function ProgramsForm(props: ProgramsFormProps) {
                 <div className="omrs-datepicker">
                   <input
                     ref={enrollmentDateRef}
+                    id="enrollmentDate"
                     type="date"
                     name="enrollmentDate"
                     max={dayjs(new Date().toUTCString()).format("YYYY-MM-DD")}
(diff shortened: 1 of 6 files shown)


src/widgets/programs/programs-form.component.tsx
```diff
@@ -228,16 +233,31 @@ export default function ProgramsForm(props: ProgramsFormProps) {
...
                   </svg>
                 </div>
+                <div className={styles.dateError}>
+                  {enrollmentDateRef &&
+                    !enrollmentDateRef?.current?.validity?.valid && (
+                      <span>
+                        <svg className="omrs-icon" role="img">
+                          <use xlinkHref="#omrs-icon-important-notification"></use>
+                        </svg>
+                        Please enter a date that is either on or before today.
```
FlorianRappl: We should use `<Trans>` here.

src/widgets/programs/programs-form.css
```diff
@@ -77,3 +77,35 @@
 .buttonStylesBorder {
   border-top: 0.125rem solid var(--omrs-color-interaction);
 }
+
+input[type="date"]:invalid {
```
FlorianRappl: This now applies to all invalid date input types. Should we set this in the `styleguide` instead?

FlorianRappl (commented):
LGTM so far!

src/widgets/programs/programs-form.css
denniskigen: Indeed we should. I've submitted a PR for this https://github.com/openmrs/openmrs-esm-styleguide/pull/75

src/widgets/programs/programs-form.component.tsx
denniskigen: Sure. Working on it.

PR #69, opened 2020-05-12
Fix NotesOverview chartBasePath route

denniskigen:
Clicking on `See All` is not leading to the user being routed to the NotesDetailedSummary. The route produced from the link does not have a chartBasePath segment. 

Trivial; Probably a typo overlooked when effecting #65.


FlorianRappl (approved):
Thanks for noticing!

PR #72, opened 2020-05-14
MF-213: Enable re-enrolling into completed programs

denniskigen:
https://issues.openmrs.org/browse/MF-213

It should be possible to re-enrol patients into programs they've already completed.

This commit fixes adds a check when determining a patient's enrolled programs so that only active enrollments are considered. Consequently, programs that a patient has unenrolled from will now be considered eligible for re-enrolment and will then appear in the Choose Program input when adding a new program.


FlorianRappl (approved):
LGTM!

PR #73, opened 2020-05-15
MF-209: Fixed links on height and weight not working

donaldkibet:
Presently the links under Results for Height and Weight bug out. This PR fixes this issue [MF-209](https://issues.openmrs.org/projects/MF/issues/MF-209)

MF-209

### DIFF
--- a/src/widgets/heightandweight/heightandweight.resource.ts
+++ b/src/widgets/heightandweight/heightandweight.resource.ts
@@ -16,9 +16,7 @@ function getDimensionsObservations(patientId: string) {
     `${fhirConfig.baseUrl}/Observation?subject:Patient=${patientId}&code=${WEIGHT_CONCEPT},${HEIGHT_CONCEPT}`
   ).pipe(
     map(({ data }) => data.entry),
-    map((entries: Array<Dimension>) =>
-      entries.map((entry: Dimension) => entry.resource)
-    ),
+    map(entries => entries.map(entry => entry.resource)),
     map(dimensions => {
       return {
         heights: dimensions.filter(dimension =>
@@ -71,7 +69,7 @@ type DimensionFetchResponse = {
   entry: Array<Dimension>;
   id: string;
   resourceType: string;
-  total: Number;
+  total: number;
   type: string;
 };
 
@@ -95,7 +93,7 @@ type Dimension = {
       type: string;
     };
     valueQuantity: {
-      value: Number;
+      value: number;
     };
   };
 };
--- a/src/widgets/heightandweight/heightandweight.resource.ts
+++ b/src/widgets/heightandweight/heightandweight.resource.ts
@@ -12,13 +12,13 @@ export function getDimensions(patientId: string) {
 }
 
 function getDimensionsObservations(patientId: string) {
-  return openmrsObservableFetch(
+  return openmrsObservableFetch<DimensionFetchResponse>(
     `${fhirConfig.baseUrl}/Observation?subject:Patient=${patientId}&code=${WEIGHT_CONCEPT},${HEIGHT_CONCEPT}`
   ).pipe(
-    map(({ data }) => {
-      return data["entry"] ? data["entry"] : [];
-    }),
-    map(entries => entries.map((entry: any) => entry.resource)),
+    map(({ data }) => data.entry),
+    map((entries: Array<Dimension>) =>
+      entries.map((entry: Dimension) => entry.resource)
+    ),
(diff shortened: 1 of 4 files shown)


src/widgets/heightandweight/heightandweight.resource.ts
```diff
@@ -15,8 +15,10 @@ function getDimensionsObservations(patientId: string) {
   return openmrsObservableFetch(
```
FlorianRappl: If we get the expected type of this request in there we don't need to be explicit with the typing below. Also we can get rid of `[]` and `any`, which give us no guarantees.

FlorianRappl (commented):
LGTM so far just improve typing and then let's merge! :rocket:

src/widgets/heightandweight/heightandweight.resource.ts
```diff
@@ -16,7 +16,7 @@ function getDimensionsObservations(patientId: string) {
     `${fhirConfig.baseUrl}/Observation?subject:Patient=${patientId}&code=${WEIGHT_CONCEPT},${HEIGHT_CONCEPT}`
   ).pipe(
     map(({ data }) => data["entry"]),
-    map((entries: []) => entries.map((entry: any) => entry.resource)),
+    map(entries => entries.map(entry => entry.resource)),
```
FlorianRappl, line 19: What I meant was that we should type the response to know exactly what `data` is. Then we know exactly whats on `data` and if we would type `entry.resourc` (missing 'e') the compiler would complain.

Right now its just empty and we need to do `data["entry"]` to go on. With `any`, which means no compiler assistance here.

donaldkibet: :+1:  thats makes it clear

jonathandick:
@FlorianRappl could you take a look at @donaldkibet 's recent changes?

src/widgets/heightandweight/heightandweight.resource.ts
```diff
@@ -64,3 +66,41 @@ function latestFirst(a, b) {
 function getDatesIssued(dimensionArray): string[] {
   return dimensionArray.map(dimension => dimension.issued);
 }
+
+type DimensionFetchResponse = {
+  entry: Array<Dimension>;
+  id: string;
+  resourceType: string;
+  total: Number;
```
FlorianRappl: The type is `number`. `Number` is the constructor.

src/widgets/heightandweight/heightandweight.resource.ts
```diff
@@ -64,3 +66,41 @@ function latestFirst(a, b) {
...
+    resourceType: string;
+    status: string;
+    subject: {
+      display: string;
+      identifier: { id: string; system: string; use: string; value: string };
+      reference: string;
+      type: string;
+    };
+    valueQuantity: {
+      value: Number;
```
FlorianRappl: Same here.

src/widgets/heightandweight/heightandweight.resource.ts
```diff
@@ -12,11 +12,13 @@ export function getDimensions(patientId: string) {
...
 
 function getDimensionsObservations(patientId: string) {
-  return openmrsObservableFetch(
+  return openmrsObservableFetch<DimensionFetchResponse>(
     `${fhirConfig.baseUrl}/Observation?subject:Patient=${patientId}&code=${WEIGHT_CONCEPT},${HEIGHT_CONCEPT}`
   ).pipe(
-    map(({ data }) => data["entry"]),
-    map((entries: []) => entries.map((entry: any) => entry.resource)),
+    map(({ data }) => data.entry),
+    map((entries: Array<Dimension>) =>
```
FlorianRappl: I think there should not be an explicit `Array<Dimension>` specified here, it should be correctly inferred by TS.

src/widgets/heightandweight/heightandweight.resource.ts
```diff
@@ -12,11 +12,13 @@ export function getDimensions(patientId: string) {
...
+      entries.map((entry: Dimension) => entry.resource)
```
FlorianRappl: Also here. Please leave that up to TS. The types should just flow correctly.

FlorianRappl (commented):
I think we are on a great way here.

Sorry for being so picky with the types. I do care a lot about correct typing, because it provides a valuable tool not only for future development, but also for maintenance and code understanding.

FlorianRappl (approved):
LGTM!

PR #75, opened 2020-05-18
Export AttachmentsOverview component

ivange94:
This was initially added in https://github.com/openmrs/openmrs-esm-patient-chart-widgets/pull/39/files#diff-fd137497c23b601a671e749b524df383R58 but somehow it's not present in master even though that PR is already merged to master. This causes the UI to fail as it cannot find the `AttachmentsOverview` component.


FlorianRappl (approved):
LGTM!

PR #76, opened 2020-05-18
MF-219: Add chart summary translations

denniskigen:

This is the beginning of efforts aimed at having full translation support across the patient chart summary.

EmptyState component:

Screenshot 2020-05-18 at 14 01 41

Widget overview components:

Screenshot 2020-05-18 at 14 39 26

### DIFF
--- a/src/ui-components/cards/summary-card-footer.component.tsx
+++ b/src/ui-components/cards/summary-card-footer.component.tsx
@@ -1,15 +1,19 @@
 import React from "react";
 import { Link } from "react-router-dom";
 import styles from "./summary-card-footer.css";
+import { Trans } from "react-i18next";
 
 export default function SummaryCardFooter(props: SummaryCardFooterProps) {
   if (!props.linkTo) {
     return (
       <div className={styles.footer}>
-        <p className="omrs-bold">See all</p>
+        <p className="omrs-bold">
+          <Trans i18nKey="See all">See all</Trans>
+        </p>
       </div>
     );
   }
+
   return (
     <div className={`${styles.footer}`}>
       <svg className="omrs-icon" fill="var(--omrs-color-ink-medium-contrast)">
@@ -20,7 +24,9 @@ export default function SummaryCardFooter(props: SummaryCardFooterProps) {
         className={`omrs-unstyled`}
         style={{ border: "none" }}
       >
-        <p className="omrs-bold">See all</p>
+        <p className="omrs-bold">
+          <Trans i18nKey="See all">See all</Trans>
+        </p>
       </Link>
     </div>
   );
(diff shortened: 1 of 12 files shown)


FlorianRappl (commented):
LGTM!

Only thing I wonder is: Should we have / adjust `en.json` here? https://github.com/openmrs/openmrs-esm-patient-chart-widgets/blob/master/translations/en.json looks like many of the things e.g., `emptyStateText` are not there yet.

denniskigen:
The interesting thing here is that it appears that the translation strings only work when they are defined in `openmrs-esm-patient-chart`. Adding them to `en.json` in this repo seems to have no effect. I don't know if this is by design or whether it is an unintentional consequence of some configuration.

FlorianRappl:
I think this is a bug potentially introduced due to the URLs only set in the app, but not here / in this module.

We should review this architecture and see if we can improve it. Having the translations in the app while they are used exclusively in widgets does not make much sense. We should put the different parts where they are used / belong to.

I'll see if I can have a look later.

src/ui-components/cards/summary-card-footer.component.tsx
```diff
@@ -1,15 +1,19 @@
...
 import styles from "./summary-card-footer.css";
+import { Trans } from "react-i18next";
 
 export default function SummaryCardFooter(props: SummaryCardFooterProps) {
   if (!props.linkTo) {
     return (
       <div className={styles.footer}>
-        <p className="omrs-bold">See all</p>
+        <p className="omrs-bold">
+          <Trans i18nKey="See all">See all</Trans>
```
ivange94, line 11: I noticed that you used the `t` function in some places while using the `Trans` component in others but I don't see much difference in context. Wouldn't `t` been enough in all instances? Could you clarify me on your reason? Not an expert with the `react-i18next` so I'm speaking based on whats on the documentation here https://react.i18next.com/latest/trans-component.

denniskigen: This is perhaps not the best use case for the `Trans` component (most of this code has been refactored locally). Basically, the `useTranslation` hook can only be used inside function components. There are occurrences in our code where helper functions have been written outside of function components and therefore `useTranslation` cannot be used. It is in these situations where I make use of the `Trans` component. 

A brief aside, you'll notice that the translation keys specified here are not specified in any standard format. I've refactored them to try as much as possible to use camelCase. I think it'd be easier going forward if we stuck to that.

ivange94: Ok. Thanks for the clarification.

denniskigen:
Closing in favour of smaller, per-widget pull requests.

PR #77, opened 2020-05-18
MF-222: Configs for vitals, weight and height metadata

nkimaina:


nkimaina:
This enables different implementors with different triage metadata like form, encounter type and concepts to use Vitals, Weight and Height features.

nkimaina:
Goes together with: https://github.com/openmrs/openmrs-esm-patient-chart/pull/160

src/config-schemas/vitals-config-schema.ts
```diff
@@ -0,0 +1,83 @@
+import { validators } from "@openmrs/esm-module-config";
+
+export const vitalsConfigSchema = {
+  vitalsEncounterTypes: {
+    arrayElements: {
+      uuid: { validators: [validators.isString] },
```
brandones: :grimacing: Do we *have to* use UUIDs? Is it possible to use something a little more human-friendly?

src/config-schemas/vitals-config-schema.ts
```diff
@@ -0,0 +1,83 @@
...
+      description: { validators: [validators.isString] }
```
brandones: Uh-oh! `description` is a reserved keyword for esm-module-config. See "[Defining a schema](https://github.com/openmrs/openmrs-esm-module-config#defining-a-schema)." What'll it be used for? Can we name the key something else?

Something I've considered is whether all the esm-module-config keywords -- `default`, `validators`, `arrayElements`, and `description`, should be prefixed, e.g. with an underscore. So config schemas would be writen like

```js
vitalsEncounterTypes: {
  _arrayElements: {
    uuid: {
      _validators: [validators.isUuid /* doesn't exist yet but should exist! */],
      _description: "An encounter type UUID"
    },
    description: {
      _validators: [validators.isString],
      _description: "Information about this type of vitals encounter"
    }
  },
  default: []
}
```

FlorianRappl: Is it a reserved prop on all levels or just for certain levels? Because on the given one I could see it as prop of the object defining `uuid`, but not in the object defining the `arrayElements`. (Just curious)

brandones: Yeah, it's reserved on all levels. I think it would be strange to make an exception here -- "description a reserved keyword unless it's the child of arrayElements, in which case it behaves like a normal config key." Seems like it would make things more confusing.

src/config-schemas/vitals-config-schema.ts
nkimaina: :laughing: I couldn't think of another way to do it that would work with REST. What do you propose?

src/config-schemas/vitals-config-schema.ts
nkimaina: sounds good. Will changes it to 'name'.

src/config-schemas/vitals-config-schema.ts
brandones: I mean, I think we should use names whenever possible. FSN for concepts. Encounter Type name for encounter types. etc. But yeah, I suppose we might be constrained by the API. @ibacher , do you have thoughts about this?

EDIT: wrote "encounter," meant "encounter type"

ibacher: `EncounterType`s always have a name which should be unique. You should be able to get it from REST like this: `/ws/rest/v1/encountertype?q=<name>`.

nkimaina: I have concerns with using this pattern for configurations i.e I would try to minimize hitting the REST endpoint unnecessarily to resolve configs. For the above scenarios, I would need to make several api calls to resolve to uuids. To me, readability of configs feels like a design/set-up time problem, and wouldn't want to tax the app later on. We actually ran into performance issues with our angular client due to many REST api calls, and had to optimize by eliminating unnecessary calls (worked perfectly on the dev environment, until it was deployed and you have users with less powerful tablets trying to use the app on a low-bandwidth 3g internet. The UX was terrible). We could try to solve the readability problem by either 1)including fields that will describe a config or 2) building tools to improve the overall experience of creating/viewing the config files.

brandones: Oh yeah, I agree, we shouldn't make *extra* calls to avoid UUIDs in the configs. It might be worth checking whether the visit data can be requested so that it comes with the encounter type name. If not though, no worries, it sounds like you're weighing this well.

translations/en.json
```diff
@@ -1,4 +1,36 @@
 {
+  "Active Medications": "Active Medications",
```
FlorianRappl: We should remove the babel plugin for automatic translation generation.

For the keys we should prefer JS identifiers.

FlorianRappl (commented):
LGTM as a whole, only thing to improve is the translation generation.

I know you did not do anything here, but you committed the change, and by now we all agreed on to remove the respective i18n babel plugin and maintain that file on our own.

FlorianRappl (approved):
LGTM!

Thanks for the work!

FlorianRappl (commented):
In general looks good so far, but you'd need to pull from master first and resolve the conflicts.

Sorry for the inconvenience!

src/widgets/vitals/vital-record.test.tsx
```diff
@@ -3,15 +3,27 @@ import { cleanup, render, wait } from "@testing-library/react";
...
+const mockUseConfig = useConfig as jest.Mock;
+const mockIsStringValidatior = validators.isString as jest.Mock;
+
+jest.mock("@openmrs/esm-module-config", () => ({
+  useConfig: jest.fn(),
+  validators: {
+    isString: jest.fn()
+  },
+  validateConfig: jest.fn()
+}));
```
brandones: You should be able to factor out all these copies of this mock into `__mocks__`

brandones: If you want :)

nkimaina: Gave it a try, wasn't successful. Could you give me pointers or hints?

brandones: Sure! It should be that all you need to do is more or less copy what's [here](https://github.com/openmrs/openmrs-esm-login/tree/master/__mocks__) and then you wire them up with moduleNameMapper in the [jest config](https://github.com/openmrs/openmrs-esm-login/blob/master/jest.config.json).

src/config-schemas/external-patient-chart-config.ts
```diff
@@ -0,0 +1 @@
+export interface ExternalPatientChartConfig {}
```
brandones: What is this, why does it exist?

nkimaina: It was to act as a contract for configs in 3rd party patient chart widgets module. It might be that you might want to enforce some structure in these configs if we are going to help load them using the configs. I don't have strong opinion about it, and might as well get rid of it.

brandones: Ah, I understand now! That's a great idea, and is something we should definitely have eventually. But I think we should wait to figure out where it should go and how it should look until we have some concrete requirement for it not to be empty.

brandones:
Uh-oh, maybe I have misunderstood something fundamental. Look at [Widget in patient-chart](https://github.com/openmrs/openmrs-esm-patient-chart/blob/master/src/view-components/widget/widget.component.tsx), specifically [these lines](https://github.com/openmrs/openmrs-esm-patient-chart/blob/af59b27e88be1a86494a72746c8b0e2c043a8272/src/view-components/widget/widget.component.tsx#L21).

There's something called `params` that is already being passed to the widgets:

```js
                  <Component
                    {...props.widgetConfig.params}
                    singleSpaContext={{ mountParcel: mountParcel }}
                  />
```

(As an aside, we should call this thing `widgetConfig.props` since it gets passed in as props, @jonathandick ).

The way to handle the widgetCommonConfig, then, is just to merge it with the widget-specific props in patient-chart.

PR #79, opened 2020-05-19
MF-231: Fix select tag styling on the programs form

denniskigen:
https://issues.openmrs.org/browse/MF-231

A custom `height` property on the `select` tag is cutting the text off on some browsers. This ticket fixes that by adjusting the `height` property so it better accommodates the text inside the `select` tag.


FlorianRappl (approved):
LGTM!

PR #80, opened 2020-05-19
MF-232: ProgramsOverview should only list active programs

denniskigen:
https://issues.openmrs.org/browse/MF-232

The ProgramsOverview component should only list active program enrollments.

This PR adds a check for active enrollments by filtering enrollments based on the value of `dateCompleted`.

### DIFF
--- a/src/widgets/programs/programs-detailed-summary.component.tsx
+++ b/src/widgets/programs/programs-detailed-summary.component.tsx
@@ -9,6 +9,7 @@ import { useTranslation } from "react-i18next";
 import { match, useRouteMatch, Link } from "react-router-dom";
 import styles from "./programs-detailed-summary.css";
 import { openWorkspaceTab } from "../shared-utils";
+import { PatientProgram } from "../types";
 
 export default function ProgramsDetailedSummary(
   props: ProgramsDetailedSummaryProps
@@ -26,7 +27,8 @@ export default function ProgramsDetailedSummary(
   useEffect(() => {
     if (patientUuid) {
       const subscription = fetchEnrolledPrograms(patientUuid).subscribe(
-        enrolledPrograms => setEnrolledPrograms(enrolledPrograms),
+        (enrolledPrograms: PatientProgram[]) =>
+          setEnrolledPrograms(enrolledPrograms),
         createErrorHandler()
       );
 
--- a/src/widgets/programs/programs-detailed-summary.component.tsx
+++ b/src/widgets/programs/programs-detailed-summary.component.tsx
@@ -14,7 +14,9 @@ import { PatientProgram } from "../types";
 export default function ProgramsDetailedSummary(
   props: ProgramsDetailedSummaryProps
 ) {
-  const [enrolledPrograms, setEnrolledPrograms] = useState(null);
+  const [enrolledPrograms, setEnrolledPrograms] = useState<
+    Array<PatientProgram>
+  >(null);
   const [
     isLoadingPatient,
     patient,
@@ -27,8 +29,7 @@ export default function ProgramsDetailedSummary(
   useEffect(() => {
     if (patientUuid) {
       const subscription = fetchEnrolledPrograms(patientUuid).subscribe(
-        (enrolledPrograms: PatientProgram[]) =>
-          setEnrolledPrograms(enrolledPrograms),
+        enrolledPrograms => setEnrolledPrograms(enrolledPrograms),
         createErrorHandler()
       );
 
--- a/src/widgets/programs/programs-form.component.tsx
+++ b/src/widgets/programs/programs-form.component.tsx
@@ -17,6 +17,7 @@ import { filter, includes, map } from "lodash-es";
 import { useHistory } from "react-router-dom";
 import { DataCaptureComponentProps } from "../shared-utils";
 import { useTranslation } from "react-i18next";
+import { LocationData, PatientProgram, Program, SessionData } from "../types";
 
 export default function ProgramsForm(props: ProgramsFormProps) {
   const formRef = useRef<HTMLFormElement>(null);
@@ -48,7 +49,7 @@ export default function ProgramsForm(props: ProgramsFormProps) {
   useEffect(() => {
     if (patientUuid && !viewEditForm) {
       const abortController = new AbortController();
-      getSession(abortController).then(({ data }) => {
+      getSession(abortController).then(({ data }: { data: SessionData }) => {
         const { sessionLocation } = data;
         if (sessionLocation && sessionLocation.uuid) {
           setLocation(sessionLocation.uuid);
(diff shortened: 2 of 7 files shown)


src/widgets/programs/programs-overview.component.tsx
```diff
@@ -4,7 +4,10 @@ import SummaryCard from "../../ui-components/cards/summary-card.component";
 import SummaryCardRow from "../../ui-components/cards/summary-card-row.component";
 import SummaryCardRowContent from "../../ui-components/cards/summary-card-row-content.component";
 import EmptyState from "../../ui-components/empty-state/empty-state.component";
-import { fetchEnrolledPrograms } from "./programs.resource";
+import {
+  fetchEnrolledPrograms,
```
FlorianRappl: Is this still needed?

src/widgets/programs/programs.resource.tsx
```diff
@@ -18,6 +18,20 @@ export function fetchEnrolledPrograms(
...
 }
 
+export function fetchActiveEnrollments(
+  patientID: string
+): Observable<PatientProgram[]> {
+  return openmrsObservableFetch(
+    `/ws/rest/v1/programenrollment?patient=${patientID}`
+  ).pipe(
+    map(({ data }) => {
+      return data["results"]
```
FlorianRappl: Why do we need to access it like this? Can we provide proper typings for the response - that we `data.results` is valid and type checked.

src/widgets/programs/programs-overview.component.tsx
denniskigen: Yes. The ProgramsDetailedSummary displays all programs and would therefore still need this.

src/widgets/programs/programs.resource.tsx
denniskigen: I've tried to address this in the latest commit.

src/widgets/programs/programs-overview.component.tsx
FlorianRappl: I meant "is the import still needed", not "is the function at all still needed" - I think it was removed now. Thanks!

src/widgets/programs/programs.resource.tsx
```diff
@@ -1,29 +1,55 @@
...
+        .sort((a, b) => (b.dateEnrolled > a.dateEnrolled ? 1 : -1))
     )
   );
 }
 
 export function getPatientProgramByUuid(
   programUuid: string
 ): Observable<PatientProgram> {
-  return openmrsObservableFetch(
+  return openmrsObservableFetch<FetchResponse<PatientProgram>>(
```
FlorianRappl: Should be just `PatientProgram`

src/widgets/programs/programs.resource.tsx
```diff
@@ -56,33 +82,28 @@ export function updateProgramEnrollment(payload, abortController) {
...
+  ).pipe(
+    map(({ data }: { data: FetchResponse<Program[]> }) => data["results"]),
+    map((results: Program[]) => results)
+  );
 }
 
-export function fetchLocations(): Observable<any> {
-  return openmrsObservableFetch(
+export function fetchLocations(): Observable<LocationData[]> {
+  return openmrsObservableFetch<FetchResponse<LocationData[]>>(
```
FlorianRappl: It's just `Array<LocationData>` (no `FetchResponse` - that's implied).

FlorianRappl (commented):
We're close! Good work :rocket:!

src/widgets/programs/programs.resource.tsx
denniskigen: Totally right, amended in the latest commit.

src/widgets/programs/programs.resource.tsx
denniskigen: Same as above, amended.

src/widgets/programs/programs-form.component.tsx
```diff
@@ -48,7 +49,7 @@ export default function ProgramsForm(props: ProgramsFormProps) {
   useEffect(() => {
     if (patientUuid && !viewEditForm) {
       const abortController = new AbortController();
-      getSession(abortController).then(({ data }) => {
+      getSession(abortController).then(({ data }: { data: SessionData }) => {
```
FlorianRappl: We don't need the type here - the response is already fully typed in `getSession`.

src/widgets/programs/programs-detailed-summary.component.tsx
```diff
@@ -26,7 +27,8 @@ export default function ProgramsDetailedSummary(
   useEffect(() => {
     if (patientUuid) {
       const subscription = fetchEnrolledPrograms(patientUuid).subscribe(
-        enrolledPrograms => setEnrolledPrograms(enrolledPrograms),
+        (enrolledPrograms: Array<PatientProgram>) =>
```
FlorianRappl: Make `useState<Array<PatientProgram>>(null)` above and drop the type here.

src/widgets/programs/programs-form.component.tsx
```diff
@@ -59,15 +60,18 @@ export default function ProgramsForm(props: ProgramsFormProps) {
 
   useEffect(() => {
     if (patientUuid) {
-      const sub1 = fetchLocations().subscribe(locations => {
-        return setLocations(locations);
-      }, createErrorHandler());
+      const sub1 = fetchLocations().subscribe(
+        (locations: Array<LocationData>) => {
```
FlorianRappl: Why do we need to specify the type? `fetchLocations` should already come with an observable of `Array<LocationData>`.

src/widgets/programs/programs.resource.tsx
```diff
@@ -1,29 +1,43 @@
...
+import { Observable } from "rxjs";
+import { map } from "rxjs/operators";
+import { FetchResponse } from "@openmrs/esm-api/dist/openmrs-fetch";
+import { PatientProgram, Program, LocationData, SessionData } from "../types";
 
 export function fetchEnrolledPrograms(
   patientID: string
-): Observable<PatientProgram[]> {
-  return openmrsObservableFetch(
+): Observable<Array<PatientProgram>> {
```
FlorianRappl: No need to type the return value - its already correct due to the `return` statement below.

As a rule of thumb: Insert as many types as necessary, but not more. Let the types flow. It will massively improve your efficiency and allow for truly agile yet totally safe refactorings.

src/widgets/programs/programs.resource.tsx
```diff
@@ -1,29 +1,43 @@
...
+    map(({ data }: { data: Array<PatientProgram> }) => data["results"]),
+    map((results: Array<PatientProgram>) =>
+      results.sort((a, b) => (b.dateEnrolled > a.dateEnrolled ? 1 : -1))
+    )
+  );
+}
+
+export function fetchActiveEnrollments(
+  patientID: string
+): Observable<Array<PatientProgram>> {
```
FlorianRappl: Again - return type is duplicated.

src/widgets/programs/programs.resource.tsx
```diff
@@ -1,29 +1,43 @@
...
+      results
+        .filter(res => !res.dateCompleted)
+        .sort((a, b) => (b.dateEnrolled > a.dateEnrolled ? 1 : -1))
     )
   );
 }
 
 export function getPatientProgramByUuid(
   programUuid: string
 ): Observable<PatientProgram> {
```
FlorianRappl: Duplicated - I would only have the type implicitly in the `return` statement if possible.

src/widgets/programs/programs.resource.tsx
```diff
@@ -56,33 +70,28 @@ export function updateProgramEnrollment(payload, abortController) {
...
-  ).pipe(map(({ data }) => data["results"]));
+  ).pipe(
+    map(({ data }: { data: Array<Program> }) => data["results"]),
+    map((results: Array<Program>) => results)
+  );
 }
 
-export function fetchLocations(): Observable<any> {
-  return openmrsObservableFetch(
+export function fetchLocations(): Observable<Array<LocationData>> {
```
FlorianRappl: Drop the type here, and leave it below.

FlorianRappl (commented):
We are on a great way here. One area to improve at the moment is types. They are super important and should not be considered "the thing I have to write". They help to maintain the code and keep the pace.

FlorianRappl (approved):
LGTM!

Actually, there are still some `data["results"]` in there instead of `data.results`, however, I would propose to have one PR to improve typings in this module.

PR #81, opened 2020-05-19
MF-230: Set Location By Default When Starting a Visit

donaldkibet:
At the moment. when you select the new visit component select location select, the location dropdown is not auto-populated with the current session default location. [MF-230](https://issues.openmrs.org/projects/MF/issues/MF-230?filter=allissues) fixes this isssue


FlorianRappl (approved):
LGTM!

PR #82, opened 2020-05-20
Updated visit location to use optional chaining

donaldkibet:
At the moment when visit location data doesn't get the location data it fails. This PR fixes this issue


drizzentic (commented):
LGTM!

src/widgets/visit/edit-visit-component.tsx
```diff
@@ -112,7 +112,7 @@ type EditVisitProps = {
 
 type PatientVisitType = {
   encounters: any[];
-  location: any;
+  location?: any;
```
FlorianRappl: Can we type these `any`s? What is the type of the `encounters` array entries? What's the type of `location`? Thanks!

FlorianRappl (commented):
Looks good to me. Typing improvements would be helpful.

FlorianRappl (approved):
LGTM! Thanks for the work :beers:!

PR #83, opened 2020-05-20
Export visit utils

nkimaina:


nkimaina:
Need to use the exported functionality in a 3rd party esm module

donaldkibet (approved):
LTGM. Should we drop the auto generated translation and manage them manually

FlorianRappl (commented):
LGTM with the exception of the translation file.

We should prefer real keys to fallbacks as keys.

That means instead of 

```
"Please enter a date that is either on or before today": "Please enter a date that is either on or before today",
```

we'd define

```
"errorFutureDate": "Please enter a date that is either on or before today",
```

nkimaina:
@FlorianRappl There's a tool that runs to build the translations json, and that's what it did when I committed my changes, I wasn't responsible for coming up with the translations. That said, I tried to change it and about 8 tests broke. On investigating, it seems to injecting the keys to the DOM, as readable text. Am I missing something? do we try to handle this issue in a separate PR?

FlorianRappl:
@nkimaina I know, but we already started removing this tool in other repositories. We should do it here, too. @brandones can give more insights!

Sure we could also do it in another PR, I'd just love to get a good grip on localizations now, because from experience it will only get harder later.

FlorianRappl (approved):

nkimaina:
No worries @FlorianRappl. Let's wait for @brandones to tell us more about what is required to get it removed because I have the mappings done correct locally.

brandones:
You can just remove it -- there's nothing to be done other than to be careful when developing and reviewing code that translation keys get added to and removed from `en.json` as necessary. But hopefully we're doing that already :)

src/widgets/visit/start-visit-prompt.ts
```diff
@@ -0,0 +1,10 @@
+import { newModalItem } from "./visit-dialog-resource";
+import { startVisitConfirmation } from "./visit-button-component";
+
+export function startVisitPrompt(onPromptClosed?: Function) {
```
FlorianRappl: For the `Function` just use `() => void`

src/widgets/visit/start-visit-prompt.ts
```diff
@@ -0,0 +1,10 @@
...
+  newModalItem({
+    component: startVisitConfirmation(),
+    name: "Prompt start Visit",
+    props: { closeComponent: () => onPromptClosed && onPromptClosed() }
```
FlorianRappl: Alternatively you can just `onPromptClosed ?.()`

FlorianRappl (commented):
Besides the few smaller cosmetic things I'd say great work!

FlorianRappl (approved):
LGTM! :rocket:

PR #84, opened 2020-05-20
MF:228 Fixed Not able to save Medication Order

donaldkibet:
At the moment, when you try to save a new or revised medication order you get an error trying saving one. This PR [MF-228](https://issues.openmrs.org/projects/MF/issues/MF-228?filter=allissues)
MF-228

### DIFF
--- a/src/widgets/medications/medication-order.component.tsx
+++ b/src/widgets/medications/medication-order.component.tsx
@@ -15,6 +15,9 @@ import { setDefaultValues, OrderMedication } from "./medication-orders-utils";
 
 const CARE_SETTINGS: string = "6f0c9a92-6f24-11e3-af88-005056821db0";
 const ORDERER: string = "e89cae4a-3cb3-40a2-b964-8b20dda2c985";
+const ORAL_ROUTE_CONCEPT: string = "160240AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
+const DAYS_DURATION_UNIT_CONCEPT: string =
+  "1072AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
 
 export default function MedicationOrder(props: MedicationOrderProps) {
   const [commonMedication, setCommonMedication] = useState([]);
@@ -26,9 +29,7 @@ export default function MedicationOrder(props: MedicationOrderProps) {
   const [dosageForm, setDosageForm] = useState("");
   const [frequencyUuid, setFrequencyUuid] = useState("");
   const [frequencyName, setFrequencyName] = useState("");
-  const [routeUuid, setRouteUuid] = useState(
-    "160240AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
-  );
+  const [routeUuid, setRouteUuid] = useState(ORAL_ROUTE_CONCEPT);
   const [routeName, setRouteName] = useState<string>(null);
   const [asNeeded, setAsNeeded] = useState(false);
   const [numRefills, setNumRefills] = useState(0);
@@ -36,17 +37,13 @@ export default function MedicationOrder(props: MedicationOrderProps) {
   const [quantity, setQuantity] = useState<string>(null);
   const [duration, setDuration] = React.useState(0);
   const [durationUnit, setDurationUnit] = React.useState(
-    "1072AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
+    DAYS_DURATION_UNIT_CONCEPT
   );
   const [durationUnitsArray, setDurationUnitArray] = useState([]);
-  const [dosingInstructions, setDosingInstructions] = useState("");
+  const [dosingInstructions, setDosingInstructions] = useState<string>();
   const [drugStrength, setDrugStrength] = useState<number>(null);
-  const [startDate, setStartDate] = React.useState(
-    dayjs(new Date()).format("YYYY-MM-DD")
-  );
-  const [endDate, setEndDate] = React.useState(
-    dayjs(new Date()).format("YYYY-MM-DD")
-  );
+  const [startDate, setStartDate] = React.useState<Date>(new Date());
+  const [endDate, setEndDate] = React.useState<Date>(new Date());
   const [
     isLoadingPatient,
     patient,
@@ -81,7 +78,7 @@ export default function MedicationOrder(props: MedicationOrderProps) {
   }, [props.drugName, patientUuid]);
 
   useEffect(() => {
-    if (startDate.length > 0 && durationUnitsArray) {
+    if (startDate && durationUnitsArray) {
       let durationPeriod = durationUnitsArray.filter(duration => {
         return duration.uuid === durationUnit;
       });
@@ -91,15 +88,17 @@ export default function MedicationOrder(props: MedicationOrderProps) {
           durationPeriod[0].display.lastIndexOf("s")
         );
         setEndDate(
-          dayjs(startDate)
-            .add(duration, durationName)
-            .format("YYYY-MM-DD")
+          new Date(
+            dayjs(startDate)
+              .add(duration, durationName)
+              .toDate()
+          )
         );
       } else {
         setEndDate(
           dayjs(startDate)
             .add(duration, "day")
-            .format("YYYY-MM-DD")
+            .toDate()
         );
       }
     }
@@ -131,7 +130,7 @@ export default function MedicationOrder(props: MedicationOrderProps) {
       getPatientDrugOrderDetails(ac, props.editProperty[0].OrderUuid).then(
         ({ data }) => {
           setEncounterUuid(data.encounter.uuid);
-          setStartDate(dayjs(data.dateActivated).format("YYYY-MM-DD"));
+          setStartDate(dayjs(data.dateActivated).toDate());
           setDosingInstructions(data.dosingInstructions);
           setDoseUnits(data.doseUnits.uuid);
           setDosageForm(data.doseUnits.display);
@@ -170,7 +169,7 @@ export default function MedicationOrder(props: MedicationOrderProps) {
     if (props.orderEdit.orderEdit) {
       const order = props.orderEdit.order;
       setEncounterUuid(order.encounterUuid);
-      setStartDate(dayjs(new Date()).format("YYYY-MM-DD"));
+      setStartDate(order.dateActivated);
       setDosingInstructions(order.dosingInstructions);
       setDoseUnits(order.doseUnitsConcept);
       setDosageForm(order.dosageForm);
@@ -376,14 +375,12 @@ export default function MedicationOrder(props: MedicationOrderProps) {
             <div className={styles.medicationOrderInput}>
               <label htmlFor="startDate">Start date</label>
               <input
-                type="Date"
+                type="date"
                 name="startDate"
                 id="startDate"
-                placeholder="Day-Month-Year"
-                autoComplete="off"
                 required
-                value={startDate}
-                onChange={evt => setStartDate(evt.target.value)}
+                value={toISODatePickerFormat(startDate)}
+                onChange={evt => setStartDate(evt.target.valueAsDate)}
               />
             </div>
             <div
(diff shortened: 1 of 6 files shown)


src/widgets/medications/medication-order.component.tsx
```diff
@@ -42,10 +42,10 @@ export default function MedicationOrder(props: MedicationOrderProps) {
   const [dosingInstructions, setDosingInstructions] = useState("");
   const [drugStrength, setDrugStrength] = useState<number>(null);
   const [startDate, setStartDate] = React.useState(
-    dayjs(new Date()).format("DD-MMM-YYYY")
+    dayjs(new Date()).format("YYYY-MM-DD")
```
FlorianRappl: Here we fix the format, but shouldn't we actually take the format from the user's locale / region?

src/widgets/medications/medication-order.component.tsx
```diff
@@ -375,13 +376,14 @@ export default function MedicationOrder(props: MedicationOrderProps) {
             <div className={styles.medicationOrderInput}>
               <label htmlFor="startDate">Start date</label>
               <input
-                type="text"
+                type="Date"
```
FlorianRappl: HTML is pretty resilient here, but we should still use the lowercase `date` here.

src/widgets/medications/medication-order.component.tsx
```diff
@@ -456,7 +458,7 @@ export default function MedicationOrder(props: MedicationOrderProps) {
             <div className={styles.medicationOrderInput}>
               <label htmlFor="endDate">End date</label>
               <input
-                type="text"
+                type="Date"
```
FlorianRappl: I know I am picky :) - sorry!

src/widgets/medications/medications-detailed-summary.component.tsx
```diff
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { Fragment } from "react";
```
FlorianRappl: Is fragment used in this code? Haven't seen it.

FlorianRappl: (I saw React.Fragment, but not Fragment); so just drop the selective import.

src/widgets/medications/medications-detailed-summary.component.tsx
```diff
@@ -32,19 +36,26 @@ export default function MedicationsDetailedSummary(
...
         medications => {
           setPastMedications(
-            medications.sort(
-              (a: any, b: any) =>
-                new Date(b.dateActivated).getDate() -
-                new Date(a.dateActivated).getDate()
-            )
+            medications
+              .sort(
+                (a: any, b: any) =>
```
FlorianRappl: Why any? What's the type?

src/widgets/medications/medications-detailed-summary.component.tsx
```diff
@@ -32,19 +36,26 @@ export default function MedicationsDetailedSummary(
...
+                  new Date(b.dateActivated).getDate() -
+                  new Date(a.dateActivated).getDate()
+              )
+              .filter((med: any) => {
```
FlorianRappl: All these anys make it hard to do refactoring and further extensions.

src/widgets/medications/medications.resource.tsx
```diff
@@ -7,14 +7,61 @@ import { Observable } from "rxjs";
...
+    dose: number;
+    doseUnits: { uuid: string; display: string };
+    dosingInstructions: string | null;
+    drug: {
+      name: string;
+      strenght: string;
+      concept: { uuid: string; display: string };
+    };
+    duration: number;
+    durationUnits: { uuid: string; display: string };
```
FlorianRappl: Some of these subtypes are quite repetitive.

Example:

```ts
interface OpenMrsReference {
  uuid: string;
  display: string;
}
```

could be used in at least 12 properties. Now if that changes then we only would need to change a single interface, instead of all occurrences in this type.

src/widgets/medications/medications.resource.tsx
```diff
@@ -33,17 +80,31 @@ export function fetchPatientMedications(
...
+  return openmrsObservableFetch<FetchResponse<PatientMedications>>(
+    `/ws/rest/v1/order?patient=${patientID}&careSetting=${CARE_SETTING}&status=${status}&v=custom:(uuid,orderNumber,accessionNumber,patient:ref,action,careSetting:ref,previousOrder:ref,dateActivated,scheduledDate,dateStopped,autoExpireDate,orderType:ref,encounter:ref,orderer:ref,orderReason,orderType,urgency,instructions,commentToFulfiller,drug:(name,strength,concept),dose,doseUnits:ref,frequency:ref,asNeeded,asNeededCondition,quantity,quantityUnits:ref,numRefills,dosingInstructions,duration,durationUnits:ref,route:ref,brandName,dispenseAsWritten)`
   ).pipe(
     map(({ data }) => {
-      const meds = [];
+      const meds: Array<PatientMedications> = [];
```
FlorianRappl, line 90: Great!

src/widgets/medications/medications.resource.tsx
```diff
@@ -33,17 +80,31 @@ export function fetchPatientMedications(
...
       data["results"].map(result => {
-        if (result.orderType.name === "Drug Order") {
+        if (result.orderType.display === "Drug Order") {
```
FlorianRappl: Can we put this in a constant somewhere? It seems so arbitrary that it may change? Not sure here. Is this not localized etc. from the API? Are we sure its always "Drug Order"?

src/widgets/medications/medications.resource.tsx
```diff
@@ -33,17 +80,31 @@ export function fetchPatientMedications(
...
+  patientID: string,
+  status: string
+): Observable<PatientMedications[]> {
+  return openmrsObservableFetch<FetchResponse<PatientMedications>>(
+    `/ws/rest/v1/order?patient=${patientID}&careSetting=${CARE_SETTING}&status=${status}&v=custom:(uuid,orderNumber,accessionNumber,patient:ref,action,careSetting:ref,previousOrder:ref,dateActivated,scheduledDate,dateStopped,autoExpireDate,orderType:ref,encounter:ref,orderer:ref,orderReason,orderType,urgency,instructions,commentToFulfiller,drug:(name,strength,concept),dose,doseUnits:ref,frequency:ref,asNeeded,asNeededCondition,quantity,quantityUnits:ref,numRefills,dosingInstructions,duration,durationUnits:ref,route:ref,brandName,dispenseAsWritten)`
+  ).pipe(
+    map(({ data }) => {
+      const meds: Array<PatientMedications> = [];
+      data["results"].map(result => {
+        if (result.orderType.display === "Drug Order") {
```
FlorianRappl: We could leverage that same constant (Drug Order) now.

FlorianRappl (commented):
I think we are on a great way here! :rocket:

src/widgets/medications/medications.resource.tsx
donaldkibet: It is a order type [Drug Order]( https://wiki.openmrs.org/display/docs/Order+Entry+API?lng=es#OrderEntryAPI-PlaceADrugOrder.1)

FlorianRappl: In the referenced document its:

```ts
"type" : "drugorder",
```

But then again - put it somewhere in a constant. We use this magic string quite some times and I would love to have some easy of maintenance and reusability.

src/widgets/medications/medication-order.component.tsx
donaldkibet: How do you recommend we use the locale format?

src/widgets/medications/medications.resource.tsx
```diff
@@ -7,14 +7,62 @@ import { Observable } from "rxjs";
...
+import { OpenmrsResource } from "../../types/openmrs-resource";
 
 const CARE_SETTING: string = "6f0c9a92-6f24-11e3-af88-005056821db0";
 const DURATION_UNITS_CONCEPT: string = "1732AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
 const NEW_MEDICATION_ACTION: string = "NEW";
 const REVISE_MEDICATION_ACTION: string = "REVISE";
 const DISCONTINUE_MEDICATION_ACTION: string = "DISCONTINUE";
-type PatientMedications = {
-  uuid: Number;
+const DRUG_ORDER_TYPE: string = "Drug Order";
```
FlorianRappl: No need to be explicit here about the type. TypeScript knows that the RHS is a `string`.

FlorianRappl: That's already the problem on the declarations above, unfortunately. We could remove the explicit type on all these assignments.

src/widgets/medications/medications.resource.tsx
```diff
@@ -33,17 +81,31 @@ export function fetchPatientMedications(
...
+  return openmrsObservableFetch<FetchResponse<PatientMedications>>(
+    `/ws/rest/v1/order?patient=${patientID}&careSetting=${CARE_SETTING}&status=${status}&v=custom:(uuid,orderNumber,accessionNumber,patient:ref,action,careSetting:ref,previousOrder:ref,dateActivated,scheduledDate,dateStopped,autoExpireDate,orderType:ref,encounter:ref,orderer:ref,orderReason,orderType,urgency,instructions,commentToFulfiller,drug:(name,strength,concept),dose,doseUnits:ref,frequency:ref,asNeeded,asNeededCondition,quantity,quantityUnits:ref,numRefills,dosingInstructions,duration,durationUnits:ref,route:ref,brandName,dispenseAsWritten)`
+  ).pipe(
+    map(({ data }) => {
+      const meds: Array<PatientMedications> = [];
+      data["results"].map(result => {
```
FlorianRappl: Can we use `data.results`? Why do we need to access it like that?

src/widgets/medications/medications.resource.tsx
```diff
@@ -33,17 +81,31 @@ export function fetchPatientMedications(
   patientID: string,
   status: string = "ACTIVE"
 ): Observable<PatientMedications[]> {
-  return openmrsObservableFetch(
-    `/ws/rest/v1/order?patient=${patientID}&careSetting=${CARE_SETTING}&status=${status}&asOfDate=${dayjs(
-      new Date()
-    ).format(
-      "YYYY-MM-DD"
-    )}&v=custom:(uuid,orderNumber,accessionNumber,patient:ref,action,careSetting:ref,previousOrder:ref,dateActivated,scheduledDate,dateStopped,autoExpireDate,orderType:ref,encounter:ref,orderer:ref,orderReason,orderType,urgency,instructions,commentToFulfiller,drug:(name,strength,concept),dose,doseUnits:ref,frequency:ref,asNeeded,asNeededCondition,quantity,quantityUnits:ref,numRefills,dosingInstructions,duration,durationUnits:ref,route:ref,brandName,dispenseAsWritten)`
+  return openmrsObservableFetch<FetchResponse<PatientMedications>>(
```
FlorianRappl: I think it should be

```ts
openmrsObservableFetch<PatientMedications>(...)
```

The result is a `Observable<FetchResponse<T>>`.

src/widgets/medications/medications.resource.tsx
```diff
@@ -33,17 +81,31 @@ export function fetchPatientMedications(
...
+      });
+      return meds;
+    })
+  );
+}
+export function fetchPatientPastMedications(
+  patientID: string,
+  status: string
+): Observable<PatientMedications[]> {
+  return openmrsObservableFetch<FetchResponse<PatientMedications>>(
```
FlorianRappl: Same problem. It's not `openmrsObservableFetch<FetchResponse<PatientMedications>>`, but rather `openmrsObservableFetch<PatientMedications>`

src/widgets/medications/medications.resource.tsx
```diff
@@ -33,17 +81,31 @@ export function fetchPatientMedications(
...
+    `/ws/rest/v1/order?patient=${patientID}&careSetting=${CARE_SETTING}&status=${status}&v=custom:(uuid,orderNumber,accessionNumber,patient:ref,action,careSetting:ref,previousOrder:ref,dateActivated,scheduledDate,dateStopped,autoExpireDate,orderType:ref,encounter:ref,orderer:ref,orderReason,orderType,urgency,instructions,commentToFulfiller,drug:(name,strength,concept),dose,doseUnits:ref,frequency:ref,asNeeded,asNeededCondition,quantity,quantityUnits:ref,numRefills,dosingInstructions,duration,durationUnits:ref,route:ref,brandName,dispenseAsWritten)`
   ).pipe(
     map(({ data }) => {
-      const meds = [];
+      const meds: Array<PatientMedications> = [];
       data["results"].map(result => {
```
FlorianRappl: `data.results`

FlorianRappl (commented):
We are on track here! :rocket:

FlorianRappl (approved):
LGTM!

PR #85, opened 2020-05-21
MF-150: Add image selection actions

ivange94:


translations/en.json
```diff
@@ -1,13 +1,47 @@
...
   "care programs": "Care Programs",
   "close": "Close",
   "conditions": "Conditions",
+  "edit": "Edit",
+  "editVisit": "Edit Visit",
+  "load": "Load",
+  "location": "Location",
   "sign": "Sign",
+  "stopDateTime": "End Date/Time",
+  "visitStartDate": "Visit Start Date",
```
FlorianRappl: We have quite a mix here between keys (`visitStartDate`) and texts (`Please enter a date that is either on or before today`). Personally, I always prefer keys. They communicate that

- it needs to be resolved properly
- what the text is about
- potentially even text specifics, e.g., is HTML allowed?

Additionally, keys are smaller in general - saving a few bytes here and there cannot hurt.

I would unify this.

FlorianRappl (commented):
Is all good - we are close. Just a few minor details.

Thanks for the work!

translations/en.json
ivange94: Actually I didn't make that change. It's done somehow automatically.

ivange94: Since I started working on this repo, when ever I make a commit or use `useTranslation` hook, the `en.json` file is modified automatically. I'd just been ignoring the changes before now.

FlorianRappl: Ok, thanks for the info - but you introduced, e.g.,

```ts
{t(
  "Attach files by dragging &amp; dropping, selecting or pasting them."
)}
```

I guess the automatic insertion is coming from a tool here. A tool that we wanted to throw out anyway. @brandones can give more details.

brandones: Yeah, the automatic insertion is done by babel-i18next-extract. As I imagine you noticed, @ivange94 , its behavior is pretty erratic. I advocate manually managing messages_en.

But, as @FlorianRappl points out, you are using `Attach files by dragging &amp; dropping, selecting or pasting them.` as a translation key. Perhaps it would be better to key it as `attachFileInstructions` or something.

ivange94: @brandones @FlorianRappl , I've updated my PR. Unfortunately I'm not able to edit `en.json`, my changes get overridden.

FlorianRappl: Hi @ivange94  as @brandones  mentions you'd need to remove the tool first.

Its in the `.babelrc` file:

```js
["i18next-extract", {
        "outputPath": "translations/{{locale}}.json",
        "discardOldKeys": true,
        "enableExperimentalIcu": true
      }]
```

If you remove these lines the tool is removed. Best case we also remove it from the `package.json`.

FlorianRappl (approved):
LGTM!

jonathandick:
@FlorianRappl, is this ready to be merged?

PR #86, opened 2020-05-21
MF-223: Fix error on End Visit action

ghost:
#MF-223
Fixed unexpected error on End visit action
Added null check fro dimensions as they can be empty
Specified jest-dom setup to be used in tests
Added unit tests for Visit Button component

### DIFF
--- a/src/widgets/visit/visit-button-component.tsx
+++ b/src/widgets/visit/visit-button-component.tsx
@@ -181,7 +181,7 @@ const cancelActiveVisit = (currentVisit: any): React.ReactNode => {
 
 const endVisit = (currentVisit: any): React.ReactNode => {
   return (
-    <div className={styles.visitPromptContainer}>
+    <div className={styles.visitPromptContainer} id="end-visit-prompt">
       <h2>Are you sure to end this visit</h2>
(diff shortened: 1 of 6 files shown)


src/widgets/visit/visit-button-component.tsx
```diff
@@ -224,9 +226,8 @@ const hideModal = () => {
   newModalItem({ component: null, name: null, props: null });
 };
 
-const visitUpdate = () => {
-  let visitData;
-  getStartedVisit.subscribe(item => (visitData = item.visitData));
+const visitUpdate = (currentVisit: any) => {
```
FlorianRappl: Can we type this? What is `any`?

FlorianRappl (commented):
Love it in general. Great tests and really well fixed. Just wanted to make sure we have great types here, too :).

Awesome job! :rocket:

src/widgets/visit/visit-button-component.tsx
ghost: Yes Sure. Will update this.

FlorianRappl (approved):
LGTM!

PR #88, opened 2020-05-27
MF 210 : Add edit options to all the results section

donaldkibet:
MF-210

- Added ability to edit vitals

### DIFF
--- a/src/widgets/heightandweight/heightandweight-record.component.tsx
+++ b/src/widgets/heightandweight/heightandweight-record.component.tsx
@@ -65,6 +65,15 @@ export default function HeightAndWeightRecord(
         <SummaryCard
           name={t("Height & Weight", "Height & Weight")}
           styles={{ width: "100%" }}
+          editComponent={VitalsForm}
+          showComponent={() =>
+            openWorkspaceTab(VitalsForm, "Edit Height and Weight", {
+              vitalUuid: dimensions?.obsData?.weight?.encounter?.reference.replace(
+                "Encounter/",
+                ""
+              )
+            })
+          }
         >
           <div className={styles.heightAndWeightContainer}>
             {!isEmpty(dimensions) && (
--- a/src/widgets/heightandweight/heightandweight-record.component.tsx
+++ b/src/widgets/heightandweight/heightandweight-record.component.tsx
@@ -67,12 +67,16 @@ export default function HeightAndWeightRecord(
           styles={{ width: "100%" }}
           editComponent={VitalsForm}
           showComponent={() =>
-            openWorkspaceTab(VitalsForm, "Edit Height and Weight", {
-              vitalUuid: dimensions?.obsData?.weight?.encounter?.reference.replace(
-                "Encounter/",
-                ""
-              )
-            })
+            openWorkspaceTab(
+              VitalsForm,
+              t("editHeightWeight", "Edit Height and Weight"),
+              {
+                vitalUuid: dimensions?.obsData?.weight?.encounter?.reference.replace(
+                  "Encounter/",
+                  ""
+                )
+              }
+            )
           }
         >
           <div className={styles.heightAndWeightContainer}>
--- a/src/widgets/vitals/vital-record.component.tsx
+++ b/src/widgets/vitals/vital-record.component.tsx
@@ -6,6 +6,8 @@ import { createErrorHandler } from "@openmrs/esm-error-handling";
 import styles from "./vital-record.css";
 import SummaryCard from "../../ui-components/cards/summary-card.component";
 import dayjs from "dayjs";
+import VitalsForm from "./vitals-form.component";
+import { openWorkspaceTab } from "../shared-utils";
 
 export default function VitalRecord(props: VitalRecordProps) {
   const [vitalSigns, setVitalSigns] = useState(null);
@@ -28,7 +30,16 @@ export default function VitalRecord(props: VitalRecordProps) {
   return (
     <>
       {vitalSigns && (
-        <SummaryCard name="Vital" styles={{ width: "100%" }}>
+        <SummaryCard
+          name="Vital"
+          styles={{ width: "100%" }}
+          editComponent={VitalsForm}
+          showComponent={() =>
+            openWorkspaceTab(VitalsForm, "Edit Vitals", {
+              vitalUuid: match.params["vitalUuid"]
+            })
+          }
+        >
           <div className={`omrs-type-body-regular ${styles.vitalCard}`}>
             <table className={styles.vitalTable}>
               <tbody>
--- a/src/widgets/vitals/vital-record.component.tsx
+++ b/src/widgets/vitals/vital-record.component.tsx
@@ -1,18 +1,23 @@
 import React, { useState, useEffect } from "react";
 import { useCurrentPatient } from "@openmrs/esm-api";
 import { useRouteMatch } from "react-router-dom";
-import { performPatientsVitalsSearch } from "./vitals-card.resource";
+import {
+  performPatientsVitalsSearch,
+  PatientVitals
+} from "./vitals-card.resource";
 import { createErrorHandler } from "@openmrs/esm-error-handling";
 import styles from "./vital-record.css";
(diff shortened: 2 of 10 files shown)


src/widgets/vitals/vitals-card.resource.tsx
```diff
@@ -6,73 +6,89 @@ import {
...
+    }),
+    map(entries => entries?.map(entry => entry.resource)),
+    map(vitals => {
       return formatVitals(
-        getVitalsByConcept(data, SYSTOLIC_BLOOD_PRESSURE_CONCEPT),
-        getVitalsByConcept(data, DIASTOLIC_BLOOD_PRESSURE_CONCEPT),
-        getVitalsByConcept(data, PULSE_CONCEPT),
-        getVitalsByConcept(data, TEMPERATURE_CONCEPT),
-        getVitalsByConcept(data, OXYGENATION_CONCEPT)
+        vitals?.filter(vital =>
```
FlorianRappl: I would drop all the `?.` here (and use `.`). Instead I would transform 

```ts
map(entries => entries?.map(entry => entry.resource))
```

to be

```ts
map(entries => entries?.map(entry => entry.resource) ?? [])
```

then we continue with an empty array.

src/widgets/vitals/vitals-card.resource.tsx
```diff
@@ -97,16 +115,18 @@ function formatVitals(
...
+      weight: weight && weight.valueQuantity.value,
+      height: height && height.valueQuantity.value
     });
   });
 }
 
-function getDatesIssued(vitalsArray): string[] {
-  return vitalsArray.map(vitals => vitals.issued);
+function getDatesIssued(vitalsArray): Array<Date> {
+  return vitalsArray?.map(vitals => vitals.issued);
```
FlorianRappl: If you *always*  return an array (e.g., via `?.map(...) ?? []`) you'd not require all the `?.` belows.

Note: Every `?.` adds to runtime. Its a couple more checks in the transformed code.

Example:

```ts
foo?.bar()
```

is usually transpiled to

```js
foo === null || foo === void 0 ? void 0 : foo.bar();
```

src/widgets/vitals/vitals-card.resource.tsx
```diff
@@ -97,16 +115,18 @@ function formatVitals(
...
 }
 
-function latestFirst(a, b) {
+function latestFirst(a: Date, b: Date) {
   return new Date(b).getTime() - new Date(a).getTime();
```
FlorianRappl, line 128: I thought `a` and and `b` are `Date`s already? Shouldn't it just be `b.getTime() - a.getTime()`?

src/widgets/vitals/vitals-detailed-summary.component.tsx
```diff
@@ -95,7 +95,7 @@ export default function VitalsDetailedSummary(
             {currentPageResults &&
               currentPageResults.map((vital, index) => {
                 return (
-                  <React.Fragment key={vital.id}>
+                  <React.Fragment key={index}>
```
FlorianRappl, line 98: Usually having an unique `id` is better than the `index`. Are we sure there will not be swap operations, additions, or removals?

What's the motivation for the change (there are sometimes good reasons to pick an `index` instead of an `id`, maybe `id` does not even exist or is not unique - just wanted to understand)?

src/widgets/vitals/vitals-overview.component.tsx
```diff
@@ -75,7 +75,7 @@ export default function VitalsOverview(props: VitalsOverviewProps) {
             <tbody>
               {currentVitals.map((vital, index) => {
                 return (
-                  <tr key={vital.id}>
+                  <tr key={index}>
```
FlorianRappl, line 78: Same:-)
Curious, here, too!

FlorianRappl (commented):
Looks great!

src/widgets/vitals/vitals-detailed-summary.component.tsx
donaldkibet: The id in this case is the encounter id which is not unique.

src/widgets/vitals/vitals-card.resource.tsx
donaldkibet: It is cast to Date but not instantiated, therefore at runtime it fails

FlorianRappl: Well, but then the provided typing is wrong.

My recommendation would be to change it to whatever it is in reality, e.g., `string` or `number`. Otherwise, instantiate beforehand.

src/widgets/vitals/vital-record.component.tsx
```diff
@@ -28,7 +30,16 @@ export default function VitalRecord(props: VitalRecordProps) {
   return (
     <>
       {vitalSigns && (
-        <SummaryCard name="Vital" styles={{ width: "100%" }}>
+        <SummaryCard
+          name="Vital"
```
FlorianRappl: Can we translate `Vital`? Sorry I know it was not part of your PR - just having a dedicated eye on localization right now.

src/widgets/vitals/vital-record.component.tsx
```diff
@@ -6,6 +6,8 @@ import { createErrorHandler } from "@openmrs/esm-error-handling";
 import styles from "./vital-record.css";
 import SummaryCard from "../../ui-components/cards/summary-card.component";
 import dayjs from "dayjs";
+import VitalsForm from "./vitals-form.component";
+import { openWorkspaceTab } from "../shared-utils";
 
 export default function VitalRecord(props: VitalRecordProps) {
   const [vitalSigns, setVitalSigns] = useState(null);
```
FlorianRappl: What's the type here?

src/widgets/vitals/vital-record.component.tsx
```diff
@@ -28,7 +30,16 @@ export default function VitalRecord(props: VitalRecordProps) {
   return (
     <>
       {vitalSigns && (
-        <SummaryCard name="Vital" styles={{ width: "100%" }}>
+        <SummaryCard
+          name="Vital"
+          styles={{ width: "100%" }}
+          editComponent={VitalsForm}
+          showComponent={() =>
+            openWorkspaceTab(VitalsForm, "Edit Vitals", {
```
FlorianRappl: Translation needed

FlorianRappl (commented):
I think we are close to finishing here.

Thanks @donaldkibet for the great work :rocket:!

src/widgets/vitals/vital-record.component.tsx
donaldkibet: `VitalsForm` is a functional component. `openWorkspaceTab` is function with return type `void`

src/widgets/heightandweight/heightandweight-record.component.tsx
```diff
@@ -65,6 +65,15 @@ export default function HeightAndWeightRecord(
         <SummaryCard
           name={t("Height & Weight", "Height & Weight")}
           styles={{ width: "100%" }}
+          editComponent={VitalsForm}
+          showComponent={() =>
+            openWorkspaceTab(VitalsForm, "Edit Height and Weight", {
```
FlorianRappl: Can we introduce a translation for "Edit Height and Weight"?

FlorianRappl (commented):
Approaching! :rocket:

src/widgets/vitals/vital-record.component.tsx
FlorianRappl: I was referring to `useState`. What type is `vitalSigns`?

(Sorry next time I'll be a bit more explicit, I thought the line number should be sufficient, but it can be ambiguous)

FlorianRappl: Is this of type `PatientVitals`?

FlorianRappl (commented):
Almost there - just this one typing enhancement (sorry, I know it was not introduced in this PR, so my intention here is just to improve the general codebase with these tiny things - thanks for your patience!).

FlorianRappl (approved):
Great work - thanks! :rocket:

PR #89, opened 2020-05-28
Consolidated use of react-router

FlorianRappl:
I've seen that we sometimes use `react-router` as import instead of `react-router-dom`.

Quite often its just typings, however, if its actual functional this is a problem. Remember that `react-router` is not shared, however,  `react-router-dom` is.

Its just a tiny fix, but may be good in the future to be included in linting, too.


brandones:
Good to know, but I'll let @jonathandick review. It's a good reminder for the whole @openmrs/microfrontends-squad .

jonathandick (approved):

PR #90, opened 2020-05-28
MF-234: Reusable RecordDetails component

denniskigen:
https://issues.openmrs.org/browse/MF-234

A pattern is being repeated across the widget record components where a summary card is being rendered as a details view. This view typically contains a table containing arbitrary data with the card header 'Details'. This logic should be extracted into its own reusable component.

Screenshot 2020-05-28 at 16 17 34

Apologies for the huge PR - some issues cropped up in the process of refactoring - for example, the logic being used to determine the `HeightAndWeight` record to load after clicking on a record in the `HeightAndWeightSummary` component wasn't working reliably - the check was returning undefined every time. I've changed it so that it compares records by their `id`s and that the individual obs would be grouped by their encounter id. This works reliably.

### DIFF
--- a/src/ui-components/cards/record-details-card.component.tsx
+++ b/src/ui-components/cards/record-details-card.component.tsx
@@ -1,0 +1,27 @@
+import React from "react";
+import SummaryCard from "../cards/summary-card.component";
+import styles from "./record-details-card.css";
+
+export default function RecordDetails(props: DetailsProps) {
+  return (
+    <SummaryCard
+      name="Details"
+      styles={{
+        width: "100%",
+        backgroundColor: "var(--omrs-color-bg-medium-contrast)"
+      }}
+    >
+      <div
+        style={props.styles}
+        className={`omrs-type-body-regular ${styles.detailsCard}`}
+      >
+        {props.children}
+      </div>
+    </SummaryCard>
+  );
+}
+
+type DetailsProps = {
+  children: React.ReactNode;
+  styles?: React.CSSProperties;
+};
--- a/src/widgets/allergies/allergy-record.component.tsx
+++ b/src/widgets/allergies/allergy-record.component.tsx
@@ -46,7 +46,7 @@ export default function AllergyRecord(props: AllergyRecordProps) {
       >
         <div
           className={`omrs-type-body-regular ${styles.allergyCard} ${
-            allergy.severity.display === "Severe"
+            allergy.severity.display === Severity.Severe
               ? `${styles.highSeverity}`
               : `${styles.lowSeverity}`
           }`}
@@ -67,16 +67,14 @@ export default function AllergyRecord(props: AllergyRecordProps) {
                 <td data-testid="severity">
                   <div
                     className={`${styles.centerItems} ${
-                      allergy.severity.display === "Severe" ? `omrs-bold` : ``
+                      styles.allergySeverity
+                    } ${
+                      allergy.severity.display === Severity.Severe
+                        ? `omrs-bold`
+                        : ``
                     }`}
-                    style={{
-                      textTransform: "uppercase",
-                      letterSpacing: "1px",
-                      lineHeight: "1rem",
-                      fontWeight: 500
-                    }}
                   >
-                    {allergy.severity.display === "Severe" && (
+                    {allergy.severity.display === Severity.Severe && (
                       <svg
                         className={`omrs-icon`}
                         fontSize={"15px"}
@@ -168,3 +166,9 @@ export default function AllergyRecord(props: AllergyRecordProps) {
 }
 
 type AllergyRecordProps = {};
+
+enum Severity {
+  Severe = "Severe",
+  Mild = "Mild",
+  Moderate = "Moderate"
+}
--- a/src/widgets/appointments/appointment-record.component.tsx
+++ b/src/widgets/appointments/appointment-record.component.tsx
@@ -9,6 +9,7 @@ import { useRouteMatch } from "react-router-dom";
 import VerticalLabelValue from "../../ui-components/cards/vertical-label-value.component";
 import styles from "./appointment-record.css";
 import { openWorkspaceTab } from "../shared-utils";
+import RecordDetails from "../../ui-components/cards/record-details-card.component";
 
 export default function AppointmentRecord(props: AppointmentRecordProps) {
   const [patientAppointment, setPatientAppointment] = useState(null);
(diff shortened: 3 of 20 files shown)


HerbertYiga:
@denniskigen   hi,thanks for the works on this,could you add a ticket ID on your description

denniskigen:
> @denniskigen   hi,thanks for the works on this,could you add a ticket ID on your description

Updated.

src/ui-components/cards/record-details-card.component.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import SummaryCard from "../cards/summary-card.component";
+import styles from "./record-details-card.css";
+
+export default function RecordDetails(props: DetailsProps) {
+  return (
+    <SummaryCard
+      name="Details"
```
FlorianRappl, line 8: Should we translate this title?

src/widgets/allergies/allergy-record.component.tsx
```diff
@@ -31,137 +32,123 @@ export default function AllergyRecord(props: AllergyRecordProps) {
...
-                      <svg
-                        className={`omrs-icon`}
-                        fontSize={"15px"}
-                        fill="var(--omrs-color-danger)"
+  return (
+    <>
+      {!!(allergy && Object.entries(allergy)) && (
+        <div className="styles.allergyContainer">
+          <SummaryCard
+            name="Allergy"
```
FlorianRappl: Should we translate here, too?

src/widgets/allergies/allergy-record.component.tsx
```diff
@@ -31,137 +32,123 @@ export default function AllergyRecord(props: AllergyRecordProps) {
...
+                    <th>Reaction</th>
+                    <th>Onset Date</th>
+                  </tr>
+                </thead>
+                <tbody>
+                  <tr>
+                    <td data-testid="severity">
+                      <div
+                        className={`${styles.centerItems} ${
+                          allergy.severity.display === "Severe"
```
FlorianRappl: Can we put the special identifier `Severe` in a common place such that it can be easily found and changed? Thanks!

src/widgets/allergies/allergy-record.component.tsx
```diff
@@ -31,137 +32,123 @@ export default function AllergyRecord(props: AllergyRecordProps) {
...
+                            ? `omrs-bold`
+                            : ``
+                        }`}
+                        style={{
+                          textTransform: "uppercase",
```
FlorianRappl: Any reason for not using a CSS file?

src/widgets/allergies/allergy-record.component.tsx
```diff
@@ -31,137 +32,123 @@ export default function AllergyRecord(props: AllergyRecordProps) {
...
+                    {/* Onset date is not yet available from the API response */}
+                    <td data-testid="onset-date">-</td>
+                  </tr>
+                </tbody>
+              </table>
+              {allergy.comment && (
+                <table className={styles.allergyTable}>
+                  <thead className="omrs-type-body-regular">
+                    <tr>
+                      <th>Comments</th>
```
FlorianRappl: Translation missing.

src/widgets/allergies/allergy-record.component.tsx
```diff
@@ -31,137 +32,123 @@ export default function AllergyRecord(props: AllergyRecordProps) {
...
+            </div>
+          </SummaryCard>
+          <RecordDetails>
+            <table
+              className={`${styles.allergyTable} ${styles.allergyDetails}`}
+            >
               <thead className="omrs-type-body-regular">
                 <tr>
-                  <th>Comments</th>
+                  <th>Last updated</th>
```
FlorianRappl: Can we have translations?

src/widgets/appointments/appointment-record.component.tsx
```diff
@@ -35,87 +36,84 @@ export default function AppointmentRecord(props: AppointmentRecordProps) {
...
+                  </td>
+                </tr>
+              </thead>
+              <tbody>
+                <tr>
+                  <td>
+                    <VerticalLabelValue
+                      label="Date"
+                      value={dayjs(patientAppointment?.startDateTime).format(
+                        "YYYY-MMM-DD"
```
FlorianRappl, line 62: This is a fixed format, however, we require a locale-specific one.

src/widgets/appointments/appointment-record.component.tsx
```diff
@@ -35,87 +36,84 @@ export default function AppointmentRecord(props: AppointmentRecordProps) {
...
+                      )}
+                      valueStyles={{ fontFamily: "Work Sans" }}
+                    />
+                  </td>
+                  <td>
+                    <VerticalLabelValue
+                      label="Start Time"
+                      value={dayjs(patientAppointment?.startDateTime).format(
+                        "HH:mm A"
```
FlorianRappl, line 71: This is a fixed format, however, we require a locale-specific one.

FlorianRappl (commented):
I think we are on a good way here.

We need to be more careful with translations. I think the need for accessing date formatting functions globally becomes stronger. We require a solution here soon. I'll look into it!

src/ui-components/cards/record-details-card.component.tsx
denniskigen: Yes, we should @FlorianRappl. I haven't found a reliable way yet to translate component prop strings that would work well with tests too. It appears things work better when I use the `Trans` component over the `useTranslation` hook. Regardless, I intend to have this resolved in my translations PR (#76). I'm already using this `RecordDetails` component there.

src/widgets/allergies/allergy-record.component.tsx
denniskigen: Is there a consensus on where this common place should be?

src/widgets/allergies/allergy-record.component.tsx
denniskigen: Don't think so - I didn't author this code originally - I'll move the styles to the CSS file.

src/ui-components/cards/record-details-card.component.tsx
FlorianRappl: Ok, thanks!

src/widgets/allergies/allergy-record.component.tsx
FlorianRappl: I don't think so - you could place it in a file like "constants.ts" or something similar. Of course, since there is no consensus we could have it in the same module, too. The only thing I would try to avoid at this point is to have the constant "disappearing" in the logic.

src/widgets/allergies/allergy-record.component.tsx
FlorianRappl: Awesome thanks! :beers:

src/widgets/allergies/allergy-record.component.tsx
denniskigen: I've moved them to an enum for now.

denniskigen:
About the missing translations, my translations PR (#76) contains a large body of work aimed at laying the groundwork for i18n. I've addressed most if not all of the translations related PR there and its close to exiting the draft-stage @FlorianRappl

src/widgets/appointments/appointment-record.component.tsx
denniskigen: How would this look like?

FlorianRappl (approved):
LGTM!

donaldkibet (approved):

PR #91, opened 2020-06-01
MF-229 Error on saving allergy

Gayatri-Kodag:
MF-229 Error on saving allergy
Added check to access valueDateTime attribute in extension array for an allergy


FlorianRappl (approved):
LGTM!

PR #92, opened 2020-06-01
MF-226 Allergy page is displayed blank

Gayatri-Kodag:
#MF-226 Allergy page is displayed blank

Updated the Environment Allergy Category to match Allergy Enum type

### DIFF
--- a/src/widgets/allergies/allergy-form.component.tsx
+++ b/src/widgets/allergies/allergy-form.component.tsx
@@ -67,7 +67,7 @@ export default function AllergyForm(props: AllergyFormProps) {
       case FOOD_ALLERGEN_CONCEPT:
         return "FOOD";
       case ENVIROMENTAL_ALLERGEN_CONCEPT:
-        return "ENVIROMENTAL";
+        return "ENVIRONMENT";
 
       default:
         "NO ALLERGEN";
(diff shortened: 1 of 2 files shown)


FlorianRappl (approved):
LGTM!

src/widgets/allergies/allergy-form.component.tsx
```diff
@@ -299,6 +299,7 @@ export default function AllergyForm(props: AllergyFormProps) {
                   name="allergenType"
                   value={ENVIROMENTAL_ALLERGEN_CONCEPT}
                   onChange={handleAllergenChange}
+                  data-testid="environmentalAllergy"
                 />
                 <span id="ENVIROMENTAL">Enviromental</span>
```
FlorianRappl, line 304: Shouldn't that id be also `ENVIRONMENT`, just for consistency?

Gayatri-Kodag: Yes, I can change the test id to `ENVIRONMENT` for maintaining the consistency. In other tests, kebab case is used for test ids. Which one should I use as test id `ENVIRONMENT` or `environment-allergen-concept`?

FlorianRappl: If its really only about styling I would agree yes. Sounds good!

(I haven't checked where the ID / element is used)

PR #93, opened 2020-06-01
MF-150: Add camera feature to attachments widget

ivange94:
Adds ability to capture attachments on the patient charts using the webcam.

### DIFF
--- a/src/widgets/attachments/attachments-overview.component.tsx
+++ b/src/widgets/attachments/attachments-overview.component.tsx
@@ -186,4 +186,4 @@ type Attachment = {
   thumbnailHeight: number;
   caption: string;
   isSelected: boolean;
-}
+};
--- a/src/widgets/attachments/attachments-overview.component.tsx
+++ b/src/widgets/attachments/attachments-overview.component.tsx
@@ -151,25 +151,24 @@ export default function AttachmentsOverview() {
   }
 
   function handleSaveImage(dataUri: string, caption: string) {
-    fetch(dataUri)
-      .then(res => res.arrayBuffer())
-      .then(buf => {
-        const file = new File([buf], caption, { type: "image/png" });
-        const abortController = new AbortController();
-        createAttachment(patientUuid, file, caption, abortController)
-          .then(res => {
-            const attachments_tmp = attachments.slice();
-            const new_att = {
-              id: `${res.data.uuid}`,
-              src: `/openmrs/ws/rest/v1/attachment/${res.data.uuid}/bytes`,
-              thumbnail: `/openmrs/ws/rest/v1/attachment/${res.data.uuid}/bytes`,
-              thumbnailWidth: 320,
-              thumbnailHeight: 212,
-              caption: res.data.comment
-            };
-          })
-          .catch(err => {});
-      });
+    const abortController = new AbortController();
+    createAttachment(patientUuid, null, caption, abortController, dataUri).then(
+      res => {
+        const attachments_tmp = attachments.slice();
+        const new_attachment = {
+          id: `${res.data.uuid}`,
+          src: `/openmrs/ws/rest/v1/attachment/${res.data.uuid}/bytes`,
+          thumbnail: `/openmrs/ws/rest/v1/attachment/${res.data.uuid}/bytes`,
+          thumbnailWidth: 320,
+          thumbnailHeight: 212,
+          caption: res.data.comment,
+          isSelected: false
+        };
+        attachments_tmp.push(new_attachment);
+        setAttachments(attachments_tmp);
+        setDataUri("");
+      }
+    );
   }
 
   return (
--- a/src/widgets/attachments/attachments-overview.component.tsx
+++ b/src/widgets/attachments/attachments-overview.component.tsx
@@ -125,7 +125,7 @@ export default function AttachmentsOverview() {
     }
   }
 
-  function handleNewAttachment(att) {
+  function handleNewAttachment(att: Attachment) {
     const attachments_tmp = attachments.slice();
     attachments_tmp.push(att);
     setAttachments(attachments_tmp);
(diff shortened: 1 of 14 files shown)


src/widgets/attachments/attachments-overview.component.tsx
```diff
@@ -124,6 +132,46 @@ export default function AttachmentsOverview() {
     }
   }
 
+  function openCamera(e: React.SyntheticEvent) {
+    setCameraIsOpen(true);
+  }
+
+  function handleTakePhoto(dataUri) {
```
FlorianRappl: `dataUri: string`?

src/widgets/attachments/attachments-overview.component.tsx
```diff
@@ -143,6 +191,24 @@ export default function AttachmentsOverview() {
             onChange={e => handleUpload(e, e.target.files)}
           />
         </form>
+        <div className={styles.cameraSection}>
+          <button className="cameraButton" onClick={openCamera}>
+            Camera
+          </button>
+          {cameraIsOpen && (
```
FlorianRappl: I think this whole thing should be moved to a new component. Ideally, the whole state related to camera is in this component.

Something like `<Camera />`. As props we'd have something like `onNewImage`, which is when a camera component informs us that a new image has been successfully uploaded and should be added to the gallery.

Would that make sense?

src/widgets/attachments/attachments-overview.css
```diff
@@ -7,3 +7,12 @@
   padding: 6px 12px;
   cursor: pointer;
 }
+
+.cameraSection {
+  padding: 10px;
+}
+
+.react-html5-camera-photo > img,
```
FlorianRappl: Are these working? I thought the class names are prefixed and do not apply.

FlorianRappl (commented):
Some minor comments.

src/widgets/attachments/attachments-overview.css
ivange94: Sorry those aren't working. That was why I created a global style and put them there. I must have forgot to delete here.

src/widgets/attachments/attachments-overview.component.tsx
ivange94: This makes sense. But do we have to name the component Camera? I mean I library already provides a component with that name. Will this act as a wrapper?

FlorianRappl: Well, first of all its about the reusability discussed last time, but second it should bring the lib into the business domain of OpenMRS. As such the API communication for upload should be part of that (as described).

The name was freely chosen. If you'd see a better name (maybe `CameraUpload`?) feel free to choose one.

src/widgets/attachments/camera-upload.component.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  function openCamera(e: React.SyntheticEvent) {
+    setCameraIsOpen(true);
+  }
+
+  function handleCloseCamera() {
+    setCameraIsOpen(false);
+  }
+
+  function handleTakePhoto(dataUri) {
```
FlorianRappl: `dataUri: string`

src/widgets/attachments/camera-upload.component.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const [dataUri, setDataUri] = useState("");
+
+  const [
+    isLoadingPatient,
+    patient,
+    patientUuid,
+    patientErr
+  ] = useCurrentPatient();
+
+  function openCamera(e: React.SyntheticEvent) {
```
FlorianRappl: `e` is not used

src/widgets/attachments/camera-upload.component.tsx
```diff
@@ -0,0 +1,81 @@
...
+        props.onNewAttachment(att);
+        setDataUri("");
+      }
+    );
+  }
+
+  return (
+    <div className={styles.cameraSection}>
+      <button className="cameraButton" onClick={openCamera}>
+        Camera
```
FlorianRappl: Translatable string.

src/widgets/attachments/image-preview.component.tsx
```diff
@@ -0,0 +1,47 @@
...
+    e.preventDefault();
+    props.onCancelCapture();
+  }
+
+  function handleSubmit(e: React.SyntheticEvent) {
+    e.preventDefault();
+    e.stopPropagation();
+  }
+
+  function updateCaption(e: React.SyntheticEvent) {
```
FlorianRappl: Prefer to use `React.ChangeEvent<HTMLInputElement>`. Then we also don't need to cast below.

src/widgets/attachments/image-preview.component.tsx
```diff
@@ -0,0 +1,47 @@
...
+    const target = e.target as HTMLInputElement;
+    setCaption(target.value);
+  }
+
+  return (
+    <form className={styles.overview} onSubmit={handleSubmit}>
+      <img src={props.dataUri} alt="Webcam preview" />
+      <input
+        type="text"
+        placeholder="Enter a caption for the image"
```
FlorianRappl: Translatable string.
(same is true for the other strings here, e.g., `Webcam preview`, `Save`, `Cancel`.

src/widgets/attachments/attachments-overview.component.tsx
```diff
@@ -124,6 +125,12 @@ export default function AttachmentsOverview() {
     }
   }
 
+  function handleNewAttachment(att) {
```
FlorianRappl: What's the type of `att`?

FlorianRappl (commented):
I think we are on a good way here!

FlorianRappl (approved):
LGTM! Great job Larry :beers:

FlorianRappl:
It seems we got in some conflicts - can you try updating / resolving them? Thanks!

cc @ivange94

ivange94:
> It seems we got in some conflicts - can you try updating / resolving them? Thanks!
> 
I just did. Should I squash the commits too?

FlorianRappl:
Its all fine - great job again!

PR #95, opened 2020-06-04
MF-239: Reusable dialog box

donaldkibet:
Screenshot from 2020-06-04 18-27-04
1. A reusable dialog-box


brandones:
This is not a widget; it is a component. A widget is a thing that is dynamically imported based on the config. Please add this directly to patient-chart. patient-chart-widgets should be only for widgets. We will need to move the other Visit* components and PatientBanner over to patient-chart as well.

jonathandick:
@brandones , I'm not sure I agree with the above. If this component is placed inside of patient-chart, and is used by a widget in patient-chart-widgets, we will create circular dependencies. We need it to be the case that chart-widgets do not import anything form patient-chart. We could put this in the style guide.

brandones:
Continuing the conversation on Slack...

FlorianRappl:
To me the given component would fit perfectly in the styleguide. From what I see there is no special domain knowledge in there.

Its a typical pattern library component. Thanks for the word @donaldkibet !

donaldkibet:
closing this PR and moving it to the style guide

PR #96, opened 2020-06-05
Fixed Non-unique key warning on height and weight 

donaldkibet:
Screenshot from 2020-06-05 13-33-48

- Fixed Nonunique keys warning.


FlorianRappl (approved):
LGTM!

PR #99, opened 2020-06-10
MF-219: Add appointments widget translations

denniskigen:

This PR:

- Adds translations strings for the components that make up the appointments widget.
- Improves the appointments widget test suite - statement coverage up to 97%.

### DIFF
--- a/src/widgets/appointments/appointments-form.component.tsx
+++ b/src/widgets/appointments/appointments-form.component.tsx
@@ -1,5 +1,5 @@
 import styles from "./appointments-form.css";
-import React, { useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import { useCurrentPatient } from "@openmrs/esm-api";
 import { getSession } from "../vitals/vitals-card.resource";
 import { createErrorHandler } from "@openmrs/esm-error-handling";
@@ -10,17 +10,17 @@ import {
 } from "./appointments.resource";
 import { useHistory } from "react-router-dom";
 import SummaryCard from "../../ui-components/cards/summary-card.component";
+import { Trans, useTranslation } from "react-i18next";
+import { DataCaptureComponentProps } from "../shared-utils";
 
-export default function AppointmentsForm() {
+export default function AppointmentsForm(props: AppointmentsFormProps) {
+  const formRef = useRef<HTMLFormElement>(null);
   const [
     isLoadingPatient,
     patient,
     patientUuid,
     patientErr
   ] = useCurrentPatient();
-  let history = useHistory();
-  let providers = null;
-  let status = null;
   const [currentSession, setCurrentSession] = useState(null);
   const [appointmentService, setAppointmentService] = useState(null);
   const [appointmentServiceType, setAppointmentServiceType] = useState(null);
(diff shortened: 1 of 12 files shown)


src/widgets/appointments/appointments-form.component.tsx
```diff
@@ -32,35 +32,58 @@ export default function AppointmentsForm() {
...
     }
 
     return () => abortController.signal;
-  }, [patientUuid]);
+  }, [patientUuid, isLoadingPatient]);
 
-  React.useEffect(() => {
+  useEffect(() => {
     const abortController = new AbortController();
     if (serviceUuid && serviceUuid != "default") {
```
FlorianRappl: Use `!==` instead of `!=`.

src/widgets/appointments/appointments-form.component.tsx
```diff
@@ -32,35 +32,58 @@ export default function AppointmentsForm() {
   const [location, setLocation] = useState(null);
   const [serviceUuid, setServiceUuid] = useState("");
   const [serviceTypeUuid, setServiceTypeUuid] = useState(null);
+  const [formChanged, setFormChanged] = useState<boolean>(false);
+  const { t } = useTranslation();
+  let history = useHistory();
+  let providers = null;
```
FlorianRappl: What are providers and status good for?

src/widgets/appointments/appointments-form.component.tsx
```diff
@@ -32,35 +32,58 @@ export default function AppointmentsForm() {
   const [location, setLocation] = useState(null);
   const [serviceUuid, setServiceUuid] = useState("");
   const [serviceTypeUuid, setServiceTypeUuid] = useState(null);
+  const [formChanged, setFormChanged] = useState<boolean>(false);
+  const { t } = useTranslation();
+  let history = useHistory();
```
FlorianRappl: Make it `const`

src/widgets/appointments/appointments-overview.component.tsx
```diff
@@ -37,36 +41,44 @@ export default function AppointmentsOverview(props: AppointmentOverviewProps) {
...
               </tr>
             </thead>
             <tbody>
-              {patientAppointments.slice(0, 5).map(appointment => {
+              {patientAppointments?.slice(0, 5).map(appointment => {
                 return (
-                  <tr key={appointment.uuid}>
-                    <td test-id="startDate">
-                      {dayjs(appointment.startDateTime).format("DD-MMM-YYYY")}
+                  <tr key={appointment?.uuid}>
```
FlorianRappl: Instead of using `appointment?.` everywhere in the map - use `filter` before the `slice` (i.e., only consider non-undefined entries), e.g.,

```ts
patientAppointments?.filter(m => !!m).slice(0, 5).map(appointment => ...)
```

FlorianRappl (commented):
I think we are on a great way here. Thanks for the PR @denniskigen ! Great work! :beers:

src/widgets/appointments/appointments-form.component.tsx
denniskigen: `providers` isn't being used anywhere, I've gotten rid of it. `status` is being used in the POST request when creating an appointment though it's still initialized to `null` at that point. I can't find docs for the Appointments resource in the REST API docs - perhaps the original author had a reason for putting it there?

src/widgets/appointments/appointments-overview.component.tsx
denniskigen: 👌

src/widgets/appointments/appointments-form.component.tsx
FlorianRappl: Ok, but so far I don't see much value in keeping `status` like that.

Can we remove this and just use `status: null` instead of `status: status`? I am easily confused, sorry.

denniskigen: Done

FlorianRappl (approved):
LGTM! Thanks for the work!

PR #100, opened 2020-06-10
MF-125: Fix timezone issues when running tests

denniskigen:
https://issues.openmrs.org/browse/MF-125

Tests that assert whether mocked times match what gets rendered in the UI sometimes fail either locally or in the CI environment because of UTC offset mismatches. Since the server uses UTC+0, it would make sense to make our local test environment use the same timezone offset to avoid such issues.

For example, the snippet attached shows an assertion that is passing locally but failing in Travis. My local environment is UTC+3 whereas Travis is UTC+0. The test failure messages show that the times are set to UTC+0 in the rendered content.

Screenshot 2020-06-10 at 21 26 02

Screenshot 2020-06-10 at 21 51 26

Screenshot 2020-06-10 at 22 06 24

I could write a test to assert that the UTC offset is being set to 0 but I'm not sure where the right place to put that would be:

`expect(new Date().getTimezoneOffset()).toBe(0)`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5591,6 +5591,58 @@
         "sha.js": "^2.4.8"
       }
     },
+    "cross-env": {
+      "version": "7.0.2",
+      "resolved": "https://registry.npmjs.org/cross-env/-/cross-env-7.0.2.tgz",
+      "integrity": "sha512-KZP/bMEOJEDCkDQAyRhu3RL2ZO/SUVrxQVI0G3YEQ+OLbRA3c6zgixe8Mq8a/z7+HKlNEjo8oiLUs8iRijY2Rw==",
+      "dev": true,
+      "requires": {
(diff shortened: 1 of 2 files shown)


donaldkibet (approved):
Ltgm

FlorianRappl (changes requested):
I am not sure here.

Real clients will also not have TZ = UTC. So its either a bug in the code that was revealed or something with the test.

This test should work **independent** of what the timezone on the client is.

denniskigen:
> I am not sure here.
> 
> Real clients will also not have TZ = UTC. So its either a bug in the code that was revealed or something with the test.
> 
> This test should work **independent** of what the timezone on the client is.

Absolutely right. This was an issue with epoch timestamps being converted to local time. Converting them to UTC date strings resolves this entirely.

Closing this on account of that resolution.

PR #101, opened 2020-06-10
MF-219: Add conditions widget translations

denniskigen:

This PR:

- Adds translations strings for the components that make up the conditions widget.
- Improves the conditions widget test suite - statement coverage up to 90%.

### DIFF
--- a/src/widgets/conditions/conditions-detailed-summary.component.tsx
+++ b/src/widgets/conditions/conditions-detailed-summary.component.tsx
@@ -1,5 +1,5 @@
 import React, { useState, useEffect } from "react";
-import { match, useRouteMatch, Link } from "react-router-dom";
+import { useRouteMatch, Link } from "react-router-dom";
 import { performPatientConditionsSearch } from "./conditions.resource";
 import { createErrorHandler } from "@openmrs/esm-error-handling";
 import { useCurrentPatient } from "@openmrs/esm-api";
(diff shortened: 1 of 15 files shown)


src/widgets/conditions/conditions-detailed-summary.component.tsx
```diff
@@ -31,28 +34,34 @@ export default function ConditionsDetailedSummary(
...
+          addComponent={ConditionsForm}
+          showComponent={() =>
+            openWorkspaceTab(ConditionsForm, `${t("Conditions Form")}`)
+          }
+        >
+          <table className={`omrs-type-body-regular ${styles.conditionTable}`}>
+            <thead>
+              <tr>
+                <td>
+                  <Trans i18nKey="condition_upper">CONDITION</Trans>
```
FlorianRappl: I would not include *upper* strings. These transformations can be made via CSS (so I would rather specify a class here that uses `text-transform: uppercase`).

src/widgets/conditions/conditions-overview.component.tsx
```diff
@@ -32,57 +33,71 @@ export default function ConditionsOverview(props: ConditionsOverviewProps) {
...
+          <SummaryCardRow>
+            <SummaryCardRowContent>
               <HorizontalLabelValue
-                label={condition.resource.code.text}
-                labelStyles={{ fontWeight: 500 }}
-                value={dayjs(condition.resource.onsetDateTime).format(
-                  "MMM-YYYY"
-                )}
-                valueStyles={{ fontFamily: "Work Sans" }}
+                label="Active Conditions"
```
FlorianRappl: Should we translate the label?

src/widgets/shared-utils.tsx
```diff
@@ -30,6 +30,11 @@ export function openWorkspaceTab<
   }
 }
 
+export const capitalize = s => {
```
FlorianRappl: Keep in mind that we have `lodash-es` as a dependency, which comes with this function.

FlorianRappl (commented):
LGTM to so - great job!

src/widgets/conditions/conditions-overview.component.tsx
denniskigen: Yes!

src/widgets/conditions/conditions-detailed-summary.component.tsx
denniskigen: Good shout, amended.

src/widgets/shared-utils.tsx
denniskigen: 🤙

FlorianRappl (approved):
LGTM! Great work @denniskigen :beers:

PR #103, opened 2020-06-12
Specify compatible patient-chart verion in readme

mddubey:
As part of this PR following things have been done:
- [x] Add Patient chart compatibility table in readme which can have a history of all the versions of patient-chart widgets and their compatible patient-chart version.
- [x] ~~Add Patient chart current version as [peer dependency](https://nodejs.org/es/blog/npm/peer-dependencies/).~~

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,9 +91,6 @@
     "react-router-dom": "^5.1.2",
     "rxjs": "^6.5.5"
   },
-  "peerDependencies": {
-    "@openmrs/esm-patient-chart-app": "0.0.1"
-  },
   "author": "",
   "license": "MPL-2.0",
   "bugs": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -91,6 +91,9 @@
     "react-router-dom": "^5.1.2",
     "rxjs": "^6.5.5"
   },
+  "peerDependencies": {
+    "@openmrs/esm-patient-chart-app": "0.0.1"
```
FlorianRappl: Seems like an anti-pattern.

Right now `@openmrs/esm-patient-chart-app` depends on `@openmrs/esm-patient-chart-widgets`, but not the other way round.

Plus we are actually working towards that the former dependency is not there.

FlorianRappl (changes requested):
I would not introduce an anti-pattern here.

package.json
mddubey: @FlorianRappl,
Thanks for your inputs. we have removed it from `peerDependencies` as discussed.

FlorianRappl (approved):
LGTM!

PR #105, opened 2020-06-13
MF-219: Add programs widget translations

denniskigen:

This PR:

- Adds translations strings for the components that make up the programs widget.
- Bumps test statement coverage up to 94%.

### DIFF
--- a/src/widgets/programs/programs-detailed-summary.component.tsx
+++ b/src/widgets/programs/programs-detailed-summary.component.tsx
@@ -1,15 +1,16 @@
 import React, { useEffect, useState } from "react";
+import { useTranslation, Trans } from "react-i18next";
+import { useRouteMatch, Link } from "react-router-dom";
 import dayjs from "dayjs";
-import SummaryCard from "../../ui-components/cards/summary-card.component";
 import ProgramsForm from "./programs-form.component";
 import { fetchEnrolledPrograms } from "./programs.resource";
 import { createErrorHandler } from "@openmrs/esm-error-handling";
 import { useCurrentPatient } from "@openmrs/esm-api";
-import { useTranslation } from "react-i18next";
-import { match, useRouteMatch, Link } from "react-router-dom";
-import styles from "./programs-detailed-summary.css";
 import { openWorkspaceTab } from "../shared-utils";
 import { PatientProgram } from "../types";
+import EmptyState from "../../ui-components/empty-state/empty-state.component";
+import SummaryCard from "../../ui-components/cards/summary-card.component";
+import styles from "./programs-detailed-summary.css";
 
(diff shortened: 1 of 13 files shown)


src/widgets/programs/programs-detailed-summary.component.tsx
```diff
@@ -27,115 +28,101 @@ export default function ProgramsDetailedSummary(
...
+                            ? `${styles.inactive}`
+                            : `${styles.active}`
+                        }`}
+                      >
+                        <td className="omrs-medium">{program.display}</td>
+                        <td>
+                          {dayjs(program.dateEnrolled).format("MMM-YYYY")}
+                        </td>
+                        <td>
+                          {program?.dateCompleted ? (
```
FlorianRappl: Inconsistent - in this branch we know (and use) program already non-conditionally, e.g., see `program.display` above. No need for the conditional member access here.

FlorianRappl (commented):
LGTM so far - thanks @donaldkibet for the great work!

Only thing I also found (but did not comment) is an inconsistent locale key use. We use snake_case (`program_title`), full sentence (`Choose a location`), and standard camelCase (`enrollmentLocation`). I would love to see a single standard that we follow.

denniskigen:
> LGTM so far - thanks @donaldkibet for the great work!
> 
> Only thing I also found (but did not comment) is an inconsistent locale key use. We use snake_case (`program_title`), full sentence (`Choose a location`), and standard camelCase (`enrollmentLocation`). I would love to see a single standard that we follow.

The agreed convention here is camelCase. This is straightforward to enforce in places where I can use the `Trans` component. Where I cannot, such as for `SummaryCard` `name`s, I use the `t` function and pass in the text content as the translation string e.g. "Choose a location". It's not ideal as it breaks the camelCase convention and is also potentially a source of confusion for someone working on a feature and wondering which approach to follow. I'd love some tips on how we can get around this.

```
<SummaryCard
  name={t("Add a new program")}
  styles={{
    width: "100%",
    backgroundColor: "var(--omrs-color-bg-medium-contrast)",
    height: "auto"
  }}
  >
    ...
 </SummaryCard>
```

FlorianRappl:
I thought the `t` function also can receive two arguments (fallback string and locale key). Wouldn't that solve the issue or did I miss something here?

denniskigen:
> I thought the `t` function also can receive two arguments (fallback string and locale key). Wouldn't that solve the issue or did I miss something here?

It can. The problem was I couldn't get access to the translated text in my tests. Only the locale keys were being rendered in the test environment so all the `TextMatch` based assertions would fail.

FlorianRappl (approved):
LGTM! Thanks for the outstanding work!

PR #106, opened 2020-06-14
Fixed Non-unique keys on height and weight

donaldkibet:
In this PR handles the following.

1. Fixed Non-Unique keys on height and weight component.
2. Improved the types on visit-note components.
3. Added edit button to vitals and height record components


FlorianRappl (approved):
LGTM! Thanks for the contribution @donaldkibet :beers:!

PR #110, opened 2020-06-16
MF-219: Add translations for reusable UI components

denniskigen:
This PR adds translations strings for the following reusable UI components:

- `HorizontalLabelValue`
- `RecordDetailsCard`
- `SummaryCard`
- `SummaryCardFooter`
- `Sidebar` 

Progress on i18n can be followed here: https://issues.openmrs.org/browse/MF-219

### DIFF
--- a/src/ui-components/cards/horizontal-label-value.component.tsx
+++ b/src/ui-components/cards/horizontal-label-value.component.tsx
@@ -1,16 +1,14 @@
 import React from "react";
 import styles from "./horizontal-label-value.css";
-import { useTranslation } from "react-i18next";
 
 export default function HorizontalLabelValue(props: HorizontalLabelValueProps) {
-  const { t } = useTranslation();
   return (
     <div className={styles.root}>
       <label
         className={props.labelClassName || "omrs-type-body-regular"}
         style={props.labelStyles}
       >
-        {t(props.label)}
+        {props.label}
         {props.specialKey && <sup>{"\u002A"}</sup>}
       </label>
       <div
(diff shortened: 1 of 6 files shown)


src/ui-components/cards/horizontal-label-value.component.tsx
```diff
@@ -1,14 +1,16 @@
...
 export default function HorizontalLabelValue(props: HorizontalLabelValueProps) {
+  const { t } = useTranslation();
   return (
     <div className={styles.root}>
       <label
         className={props.labelClassName || "omrs-type-body-regular"}
         style={props.labelStyles}
       >
-        {props.label}
+        {t(props.label)}
```
FlorianRappl: I think this is dangerous. The user of the component does not know that the input will be localized. I think that all props are expected as localized. Otherwise, we'll have a problem.

Example: If React would do what we do here now, we would only supply `<div>my text</div>`, which is then essentially equivalent to `<div><Trans>my text</Trans></div>`, however, what if we already supplied (by not knowing) a translated string? It would be `<div><Trans><Trans>my text</Trans></Trans></div>`. The only way out of this is to state "this component takes the input as-is - if you want it translated, then please translate".

src/ui-components/cards/summary-card-footer.component.tsx
```diff
@@ -1,17 +1,20 @@
...
-        <p className="omrs-bold">See all</p>
+        <p className="omrs-bold">
+          <Trans i18nKey="seeAll">See all</Trans>
+        </p>
       </div>
     );
   }
   return (
-    <div className={`${styles.footer}`}>
+    <div className={styles.footer}>
```
FlorianRappl, line 17: :heart:

src/ui-components/cards/horizontal-label-value.component.tsx
denniskigen: I agree. I've amended it so that it receives already translated strings in the latest commit.

FlorianRappl (approved):
LGTM!

Thanks for the work @denniskigen !

PR #112, opened 2020-06-16
MF:245 Fixed save button Under Notes To be enabled when valid entry is made

donaldkibet:
1. This PR fixes the this issue with save button on notes [MF-245](https://issues.openmrs.org/projects/MF/issues/MF-245?filter=allissues).
2. Fixed Non-unique keys on search by diagnosis result response.


FlorianRappl (approved):
LGTM!

PR #113, opened 2020-06-17
MF-150: Add user authorization to attachments widget

ivange94:


FlorianRappl:
The build is failing - I guess we need to pull from `master` first?
cc @ivange94

src/widgets/attachments/attachments-overview.component.tsx
```diff
@@ -9,6 +9,7 @@ import Gallery from "react-grid-gallery";
 import styles from "./attachments-overview.css";
 import CameraUpload from "./camera-upload.component";
 import { Trans } from "react-i18next";
+import { UserHasAccessReact } from "@openmrs/esm-api";
```
mks-d: @FlorianRappl @jonathandick I find this naming weird: [`UserHasAccessReact`](https://github.com/openmrs/openmrs-esm-api/blob/7f0f30916edc18644c72b5f2ee0a345acdd4d207/src/shared-api-objects/user-has-access-react.component.tsx#L4), why "react"?

src/widgets/attachments/attachments-overview.component.tsx
```diff
@@ -131,50 +132,54 @@ export default function AttachmentsOverview() {
...
-      </div>
-      {getSelectedImages().length !== 0 && (
-        <div className={styles.actions}>
-          <button
-            onClick={deleteSelected}
-            className={`omrs-btn omrs-filled-action`}
-          >
-            <Trans i18nKey="deleteSelected">Delete selected</Trans>
-          </button>
+    <UserHasAccessReact privilege="View Attachments">
```
mks-d, line 134: @jonathandick so @ivange94 is suggesting that we introduce a pair view + edit privileges tentatively named:
* `"View Attachments"`
* `"Create Attachments"`

But we need to come up with some sort of convention here right?

mks-d: I like the pair view + edit because this sticks to the OpenMRS data model, are we ok to try to stick to this across MF?

mks-d (commented):
@jonathandick a couple of questions here around privileges.

src/widgets/attachments/attachments-overview.component.tsx
FlorianRappl: That is not my choice, but I guess the intention was to signal that the given component is actually a React component.

FlorianRappl (approved):
LGTM! Thanks for the great job @ivange94 !

PR #114, opened 2020-06-22
Fixed issue with condition navigation to conditions

donaldkibet:
At the moment when you try to view the condition record, you are unable to view it. This PR fixes this bug [MF-223](https://issues.openmrs.org/browse/MF-233)


FlorianRappl (approved):
LGTM! :+1:

PR #115, opened 2020-06-25
MF-222  Create configurations for vitals, height and weight widgets

brandones:
https://issues.openmrs.org/browse/MF-222

Uses https://github.com/openmrs/openmrs-esm-module-config/pull/49

As demonstrated in today's meeting. Supports a config like

```js
  "@openmrs/esm-patient-chart-app": {
    dashboardDefinitions: [
      {
        name: "summaryDashboard",
        title: "Overview",
        layout: { columns: 4 },
        widgets: [
          {
            name: "VitalsOverview",
            esModule: "@openmrs/esm-patient-chart-widgets",
            layout: { columnSpan: 2 },
            props: { basePath: "results/vitals" },
            config: { concepts: {
              oxygenSaturationUuid: "5088AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" }
            }
          }
        ]
      }
    ]
  },
  "@openmrs/esm-patient-chart-widgets": {
    concepts: {
      diastolicBloodPressureUuid: "5085AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
    }
  }
```

Where values provided via the dashboard config override the values provided globally to the widget module.


README.md
```diff
@@ -13,3 +13,32 @@ More info on the [wiki][]
...
+
+## Configuration
+
+<!-- GENERATED BY OPENMRS CONFIG CLI -->
+See the [openmrs-esm-module-config docs](https://wiki.openmrs.org/display/projects/openmrs-esm-module-config)
+for information about how to provide configuration files.
+
+- concepts
+  - **systolicBloodPressureUuid** *(default: 5085AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA)*\
+    *.*
```
brandones: This should be fixed by https://github.com/openmrs/openmrs-esm-cli/pull/7

__mocks__/vitals.mock.ts
```diff
@@ -827,7 +827,7 @@ export const mockVitalData = [
     diastolic: 72,
     pulse: 22,
     temperature: 37,
-    oxygenation: 30
+    oxygenSaturation: 30
```
brandones, line 830: Pedantry unrelated to the topic of this PR: I standardized all the vitals vocabulary to match CIEL preferred terms.

src/widgets/vitals/vital-record.test.tsx
```diff
@@ -71,21 +71,21 @@ describe("<VitalRecord />", () => {
 
     await wait(() => {
       expect(wrapper).toBeDefined();
```
brandones, line 73: If you haven't read it, I *highly* recommend [Common Mistakes with React Testing Library](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library)

src/widgets/vitals/vitals-card.resource.tsx
```diff
@@ -155,60 +152,29 @@ export function savePatientVitals(
...
-      obs: isVitalValid(vitals)
+      encounterType: encounterTypeUuid,
+      form: formUuid,
+      obs: createObsObject(vitals, concepts)
     }
   });
 }
 
-function isVitalValid(vitals: Vitals): any[] {
+function createObsObject(
```
brandones, line 162: Please give functions names that explain what they do

jonathandick (approved):
This is outstanding work.

__mocks__/vitals.mock.ts
FlorianRappl: Love it!

package.json
```diff
@@ -27,7 +28,7 @@
   ],
   "husky": {
     "hooks": {
-      "pre-commit": "pretty-quick --staged && concurrently npm:lint npm:typescript npm:coverage"
+      "pre-commit": "pretty-quick --staged && concurrently npm:lint npm:typescript npm:coverage npm:document && git add README.md"
```
FlorianRappl: Yeah I see what you are doing here, but I also have one concern: If I / somebody has changes in README.md that should *not* be committed, this one will stage them (and commit them).

Right now I don't know any way around (of course skipping the hook is possible, but usually the case above occurs when people do not think about it; otherwise these README.md changes would be stashed).

From my point of view we can leave it as changed, but I would either communicate that or look for a refinement.

src/widgets/vitals/vital-record.test.tsx
FlorianRappl: I agree, but I also think in the given context it does not matter much. Its not pretty (nor efficient), but in the end we should not forget why we write tests - to actually *test* things (and not to have the prettiest / most efficient code - which this one obviously is not; independent of `wait` or not).

FlorianRappl: Maybe we should do a React dev session on testing here? Are you willing to spend an hour with the team @brandones ?

src/widgets/vitals/vitals-card.resource.tsx
FlorianRappl: But I would also love to see a great return type. Can we improve the `any[]` to `Array<TheRealType>`?

```ts
interface ObservationRecord {
  concept: string;
  value: string;
}
```

**Remark**: I guess you have a much better name here, also I am not sure if the value is indeed a `string`, I guess for most concepts I see below `number` is better. Not sure what's returned from the API as we have, unfortunately, not typed it properly.

FlorianRappl (commented):
LGTM so far!

Only improvement I see (but we can also separate that) is the typing area.

src/widgets/vitals/vital-record.test.tsx
brandones: Actually, right here it makes the difference between a good failure message and a useless one. Before, the error message it produced was something like "timed out before condition met" -- it could have been any of those problems. Now it tells you which condition failed.

brandones: Yeah, that might be fun! Though honestly everything I know and would have to share is just stuff I've read on Kent C Dodds's blog.

FlorianRappl: Yeah that's what I would call a cosmetic improvement. The functional one (test is green or red) is not impacted (yet). Nevertheless, as already written - its better, even though I'd go with correct typings and an intrinsic correctness before excessive testing any time.

package.json
brandones: Makes sense. I will think about how to improve this workflow. Removing the auto-add for now.

src/widgets/vitals/vitals-card.resource.tsx
brandones: Good suggestion, thanks!

FlorianRappl (approved):
Awesome work! Thanks @brandones  :beers:

src/widgets/vitals/vital-record.test.tsx
brandones: > The functional one (test is green or red) is not impacted (yet).

Not sure I understand. Maybe it would clarify to say what you would change about this test?

> cosmetic improvement

Sure, but I think the importance of good error messages is often overlooked. 80% of coding is debugging. Having high-quality error messages is a serious quality of life issue for a programmer.

FlorianRappl: > Not sure I understand. Maybe it would clarify to say what you would change about this test?

What I'm saying is that a test is either OK (green) or broken (red). This is the purpose of the test. When this behavior did not change (I guess it did not) the functional behavior remains unchanged. At the end, this is what matters most.

> Sure, but I think the importance of good error messages is often overlooked. 80% of coding is debugging. Having high-quality error messages is a serious quality of life issue for a programmer.

I think that is undeniably true (well, the 80% we can argue - but let's say a large fraction of dev certainly is), but if there is an issue with the test (i.e., test is now red) you'll need to start debugging anyway. Nevertheless, having a good (or more appropriate) message certainly helps, so again - this is appreciated. :+1:

PR #116, opened 2020-07-02
MF-150: Add option to edit caption and also show attachment timestamps

ivange94:
Edit caption by clicking on the image captions. 
Screen Shot 2020-07-02 at 11 54 44

### DIFF
--- a/src/widgets/attachments/attachments-overview.component.tsx
+++ b/src/widgets/attachments/attachments-overview.component.tsx
@@ -10,6 +10,7 @@ import styles from "./attachments-overview.css";
 import CameraUpload from "./camera-upload.component";
 import { Trans } from "react-i18next";
 import AttachmentThumbnail from "./attachment-thumbnail.component";
+import dayjs from "dayjs";
 
 export default function AttachmentsOverview() {
   const [attachments, setAttachments] = useState<Attachment[]>([]);
@@ -35,7 +36,7 @@ export default function AttachmentsOverview() {
             thumbnailHeight: 212,
             caption: attachment.comment,
             isSelected: false,
-            dateTime: `${new Date(attachment.dateTime)}`
+            dateTime: dayjs(attachment.dateTime).format("YYYY-MM-DD HH:mm:ss")
           }));
           setAttachments(listItems);
         }
@@ -61,7 +62,9 @@ export default function AttachmentsOverview() {
                 thumbnailHeight: 212,
                 caption: response.data.comment,
                 isSelected: false,
-                dateTime: `${new Date(response.data.dateTime)}`
+                dateTime: dayjs(response.data.dateTime).format(
+                  "YYYY-MM-DD HH:mm:ss"
+                )
               };
               attachments_tmp.push(new_attachment);
             }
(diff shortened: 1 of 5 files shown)


src/widgets/attachments/attachments-overview.component.tsx
```diff
@@ -33,7 +34,8 @@ export default function AttachmentsOverview() {
             thumbnailWidth: 320,
             thumbnailHeight: 212,
             caption: attachment.comment,
-            isSelected: false
+            isSelected: false,
+            dateTime: `${new Date(attachment.dateTime)}`
```
FlorianRappl: What's the difference between `new Date(isoDateHere).toISOString()` and just `isoDateHere`?

src/widgets/attachments/attachments-overview.component.tsx
```diff
@@ -9,6 +9,7 @@ import Gallery from "react-grid-gallery";
 import styles from "./attachments-overview.css";
 import CameraUpload from "./camera-upload.component";
 import { Trans } from "react-i18next";
+import AttachmentThumbnail from "./attachment-thumbnail.component";
 
 export default function AttachmentsOverview() {
   const [attachments, setAttachments] = useState([]);
```
FlorianRappl: Can you specify the type here?

FlorianRappl (commented):
Looks good so far - thanks @ivange94 !

src/widgets/attachments/attachments-overview.component.tsx
ivange94: @FlorianRappl not sure I understand the question. But the reason I created a `Date` object from the `dateTime` value is because the `dateTime` value was not very readable. This is a sample `2020-04-20T13:18:57.000+0100`.

FlorianRappl: When I do

```js
`${new Date()}`
// "Thu Jul 09 2020 08:06:41 GMT+0200 (Central European Summer Time)"
```

it is not so readable either. Furthermore, it presents a potential problem with the chosen locale. Could we use something like `dayjs` (used anyway in the lib) to format the date properly? Many thanks!

ivange94: > Could we use something like dayjs (used anyway in the lib) to format the date properly? 

I couldn't find any reference to `dayjs` in the [library](https://github.com/benhowell/react-grid-gallery/search?q=DayJS&unscoped_q=dayjs). And it seems it's not currently been used in the widgets repo either. I wanted to confirm that I understand what you meant before I introduce a new dependency.

FlorianRappl: Maybe I'm blind:
https://github.com/openmrs/openmrs-esm-patient-chart-widgets/blob/master/package.json#L87

Just use this one - we use it in other microfrontends, too, for formatting and computations on date and time objects.

And sorry for the confusion, when I wrote "used anyway in the lib" I mean of course "used anyway in the current package", where lib = current package = openmrs-patient-chart-widgets. I thought that was clear. My mistake.

ivange94: @FlorianRappl I'm not sure how I missed this. Sorry about that. Please let me know if this is ok now. Visually it looks better than before.

jonathandick (approved):
LGTM

PR #117, opened 2020-07-02
Fix withConfig so that vitals form works again

brandones:
So I think what was happening is that before, the inner component was being re-created on each run of the outer component. Besides being inefficient, this caused the wrapped component state to get wiped out on each rerender.

I fixed it by making the code more normative -- I expressed the complex HOC as a composition of two very normal HOCs. Best practices FTW.

Screenshot from 2020-07-02 11-32-14

### DIFF
--- a/src/with-config.tsx
+++ b/src/with-config.tsx
@@ -11,7 +11,7 @@ export default function withConfig(Comp) {
 }
 
 function provideContext(Comp) {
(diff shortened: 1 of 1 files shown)


jonathandick (approved):

denniskigen:
Works great locally for me. Thanks for the quick work @brandones!

PR #118, opened 2020-07-07
MF-247 | Implementation of Immunization Widget

mddubey:
As part of this PR below things are done:
- [x] Introduce a new widget to capture and display Immunization for a patient.
- [x] Introduce a summary component for the Immunization. 
- [x] Change the Visit screen and patient banner components to translations if provided.


src/widgets/immunizations/immunization-mapper.ts
```diff
@@ -0,0 +1,111 @@
...
+  const sequenceLabel = protocolApplied?.series;
+  const sequenceNumber = protocolApplied?.doseNumberPositiveInt;
+  const occurrenceDateTime = dayjs(
+    immunizationResource?.resource?.occurrenceDateTime
+  ).format("YYYY-MM-DD");
+  const expirationDate = dayjs(
+    immunizationResource?.resource?.expirationDate
+  ).format("YYYY-MM-DD");
+  return {
+    encounterUuid: encounterUuid,
```
FlorianRappl: Prefer shorthand if possible

```
return {
  encounterUuid,
  immunizationObsUuid,
  // ...
};
```

src/widgets/immunizations/immunization-mapper.ts
```diff
@@ -0,0 +1,111 @@
...
+    manufacturer: manufacturer,
+    lotNumber: lotNumber,
+    sequenceLabel: sequenceLabel,
+    sequenceNumber: sequenceNumber,
+    occurrenceDateTime: occurrenceDateTime,
+    expirationDate: expirationDate
+  };
+};
+
+export const mapFromFhirImmunizationSearchResults = immunizationSearchResult => {
```
FlorianRappl: Can we type this? What is `immunizationSearchResult`?

src/widgets/immunizations/immunization-mapper.ts
```diff
@@ -0,0 +1,111 @@
+import { get, groupBy, map, orderBy } from "lodash-es";
+import dayjs from "dayjs";
+
+const mapToImmunizationDoses = immunizationResource => {
```
FlorianRappl: Can we type this? What is `immunizationResource`?

src/widgets/immunizations/immunization-mapper.ts
```diff
@@ -0,0 +1,111 @@
...
+      vaccineName:
+        immunizationResources[0]?.resource?.vaccineCode?.coding[0].display,
+      vaccineUuid:
+        immunizationResources[0]?.resource?.vaccineCode?.coding[0].code,
+      doses: orderBy(doses, [dose => get(dose, "occurrenceDateTime")], ["desc"])
+    };
+  });
+};
+
+export const mapToFhirImmunizationResource = (
```
FlorianRappl: Missing types on all parameters.

src/widgets/immunizations/immunization-mapper.ts
```diff
@@ -0,0 +1,111 @@
...
+  code: string;
+  system: string;
+  display: string;
+};
+
+type FHIRImmunizationResource = {
+  resourceType: "Immunization";
+  status: "Completed";
+  id: string;
+  vaccineCode: { coding: [Code] };
```
FlorianRappl: Just to be sure - this should be a tuple with 1 entry? Or did you mean do have `Array<Code>` here?

src/widgets/immunizations/immunization-widget-config-schema.ts
```diff
@@ -0,0 +1,60 @@
...
+          }
+        ]
+      }
+    ]
+  }
+};
+
+type ImmunizationSequenceDefinition = {
+  vaccineConceptUuid: string;
+  sequences: Array<object>;
```
FlorianRappl: Can we be more exact than `object`?

src/widgets/immunizations/immunizations-form.component.tsx
```diff
@@ -0,0 +1,348 @@
...
+        immunizationObsUuid,
+        vaccineName,
+        vaccineUuid,
+        manufacturer,
+        expirationDate,
+        vaccinationDate,
+        lotNumber,
+        sequences,
+        currentDose
+      }: Immunization = props.match.params[0];
```
FlorianRappl: Does this work? Usually the `match` params are always strings. Also I don't know if you can access them by number - usually they use their id, e.g., in a route definition `/foo/:name` the params of `/foo/flo` would be `{ name: 'flo' }`.

Maybe this is something different and I don't understand what `match` should be in this case (or `ImmunizationsFormProps` in general). Can we type this better?

src/widgets/immunizations/immunizations-form.component.tsx
```diff
@@ -0,0 +1,348 @@
...
+  }, [props.match.params]);
+
+  const handleFormSubmit = event => {
+    event.preventDefault();
+    const currentVisitUuid = getStartedVisit?.getValue()?.visitData?.uuid;
+    const currentLocationUuid = currentUser?.sessionLocation?.uuid;
+    const currentProviderUuid = currentUser?.currentProvider?.uuid;
+
+    const immunization: Immunization = {
+      patientUuid: patientUuid,
```
FlorianRappl: Please use shorthands if possible

src/widgets/immunizations/immunizations-form.component.tsx
```diff
@@ -0,0 +1,348 @@
...
+
+  function navigate() {
+    history.push(`/patient/${patientUuid}/chart/immunizations`);
+    props.closeComponent();
+  }
+
+  function createForm() {
+    const addFormHeader = t("add vaccine", "Add Vaccine") + ": " + vaccineName;
+    const editFormHeader =
+      t("edit vaccine", "Edit Vaccine") + ": " + vaccineName;
```
FlorianRappl: The `vaccineName` should rather be an argument for the localization. The `{label}: {value}` is not applicable in all languages, in some you may want something different from a colon, in others you may want to switch order.

src/widgets/immunizations/immunizations-form.component.tsx
```diff
@@ -0,0 +1,348 @@
...
+    return () => abortController.abort();
+  };
+
+  function navigate() {
+    history.push(`/patient/${patientUuid}/chart/immunizations`);
+    props.closeComponent();
+  }
+
+  function createForm() {
+    const addFormHeader = t("add vaccine", "Add Vaccine") + ": " + vaccineName;
```
FlorianRappl: The `vaccineName` should rather be an argument for the localization. The `{label}: {value}` is not applicable in all languages, in some you may want something different from a colon, in others you may want to switch order.

src/widgets/immunizations/immunizations-form.component.tsx
```diff
@@ -0,0 +1,348 @@
...
+                      <option value="DEFAULT">
+                        {t("please select", "Please select")}
+                      </option>
+                      {immunizationSequences.map(s => {
+                        return (
+                          <option
+                            key={s.sequenceNumber}
+                            value={s.sequenceNumber}
+                          >
+                            {t(s.sequenceLabel, s.sequenceLabel)}
```
FlorianRappl, line 186: Are we sure we can translate `sequenceLabel`? Is this coming from the API (I guess so)?

src/widgets/immunizations/immunizations-form.component.tsx
```diff
@@ -0,0 +1,348 @@
...
+        </div>
+      </form>
+    );
+  }
+
+  const closeForm = event => {
+    let userConfirmed: boolean = false;
+    if (formChanged) {
+      userConfirmed = confirm(
+        "There is ongoing work, are you sure you want to close this tab?"
```
FlorianRappl: Well, I don't think we should use `confirm`, but that left aside we need a general way of dealing with that. Usually, you use a feature of (all known to me) routers; the ability to block transitions. Additionally, you'd need to hook on to the unload event to prevent closing the website.

We should have a convenience function (or hook) that we use for any form.

Independent of the technical implementation here I'd still ask for a translation of the string. Thanks!

src/widgets/immunizations/immunizations-form.component.tsx
```diff
@@ -0,0 +1,348 @@
...
+      history.push(`/patient/${patientUuid}/chart/immunizations`);
+      props.closeComponent();
+    } else if (!formChanged) {
+      props.entryCancelled();
+      history.push(`/patient/${patientUuid}/chart/immunizations`);
+      props.closeComponent();
+    }
+  };
+
+  const onDoseSelect = event => {
```
FlorianRappl: Can you put that up or make it a `function` to use hoisting here? Thanks! (also either type the `event` or use it as an anonymous lambda above, where `event` would be inferred).

src/widgets/immunizations/immunizations-overview.component.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [patientImmunizations, setPatientImmunizations] = useState(null);
+  const [
+    isLoadingPatient,
+    patient,
+    patientUuid,
+    patientErr
+  ] = useCurrentPatient();
+  const { t } = useTranslation();
+  const chartBasePath = useChartBasePath();
+  const immunizationsPath = chartBasePath + "/" + props.basePath;
```
FlorianRappl: A template string would make sense here.

src/widgets/immunizations/vaccination-row.component.tsx
```diff
@@ -0,0 +1,194 @@
...
+              </svg>
+            </Link>
+          }
+        </td>
+      </tr>
+    );
+  });
+}
+
+const capitalize = s => {
```
FlorianRappl: We could use the same function from `lodash`. I think `lodash` is anyway used in the lib, right?

FlorianRappl: Otherwise, please type it properly ;)

src/widgets/immunizations/vaccination-row.component.tsx
```diff
@@ -0,0 +1,194 @@
...
+  if (typeof s !== "string") return "";
+  return s.toUpperCase();
```
FlorianRappl: Hm I thought you wanted to transform, e.g., `foo` to `Foo`, but this function transforms `foo` to `FOO`. Really wanted? Why? You could also use a CSS transform for that.

src/widgets/immunizations/vaccination-row.component.tsx
```diff
@@ -0,0 +1,194 @@
...
+    return (
+      recentDose.sequenceLabel +
+      " on " +
+      dayjs(recentDose.occurrenceDateTime).format("DD-MMM-YYYY")
+    );
+  }
+  const singleDoseText = t("single dose", "Single Dose");
+  return (
+    singleDoseText +
+    " on " +
```
FlorianRappl: Actually I would not do translation like this. Instead, I would add an argument "date" to the translation string and bring this one in dynamically. That way the exact location (and connection word, e.g., in English "on") would be left to the translation.

src/widgets/immunizations/vaccination-row.component.tsx
```diff
@@ -0,0 +1,194 @@
...
+    dayjs(recentDose.occurrenceDateTime).format("DD-MMM-YYYY")
+  );
+}
+
+function hasExistingDoses(patientImmunization: any) {
```
FlorianRappl: You guess what's coming... What is `any`? Can we be more exact? It seems like it is something

```ts
patientImmunization: { doses: Array<T> }
```

where we don't really care about `T`.

src/widgets/immunizations/vaccination-row.component.tsx
```diff
@@ -0,0 +1,194 @@
...
+  return patientImmunization.doses && patientImmunization.doses.length > 0;
+}
+
+function hasSequence(patientImmunization) {
```
FlorianRappl: What is `patientImmuninization`?

src/widgets/immunizations/vaccination-row.component.tsx
```diff
@@ -0,0 +1,194 @@
...
+  return (
+    patientImmunization?.sequences && patientImmunization?.sequences?.length > 0
+  );
+}
+
+function renderSequenceTable(match, t, immunization) {
```
FlorianRappl: What are these parameters?

src/widgets/immunizations/vaccination-row.component.tsx
```diff
@@ -0,0 +1,194 @@
...
+              </table>
+            </td>
+          </tr>
+        )}
+      </React.Fragment>
+    )
+  );
+}
+
+function getRecentVaccinationText(t, patientImmunization) {
```
FlorianRappl: Types of these parameters?

src/widgets/immunizations/immunizations.resource.tsx
```diff
@@ -0,0 +1,78 @@
...
+  patientIdentifier: string,
+  patientUuid: string,
+  abortController: AbortController
+) {
+  return openmrsFetch(`/ws/rest/v1/${patientUuid}/fhir/immunization`, {
+    signal: abortController.signal
+  }).then(response => response.data);
+}
+
+export function savePatientImmunization(
```
FlorianRappl, line 66: Missing parameter types

src/widgets/immunizations/immunizations-form.component.tsx
```diff
@@ -0,0 +1,348 @@
...
+  entrySubmitted: () => {},
+  closeComponent: () => {}
+};
+
+function hasSequences(sequences) {
+  return sequences && sequences?.length > 0;
+}
+
+type ImmunizationsFormProps = DataCaptureComponentProps & {
+  match: any;
```
FlorianRappl: We should improve this type here!

src/widgets/immunizations/immunizations-detailed-summary.component.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  function findExistingDoses(existingImmunizationsForPatient) {
+    return immunizationFromConfig => {
+      const consolidatedImmunization = {
+        vaccineName: immunizationFromConfig.display,
+        vaccineUuid: immunizationFromConfig.uuid,
+        sequences: immunizationFromConfig.sequences,
+        doses: []
+      };
+      const isSameImmunization = existingImmunization =>
```
FlorianRappl: Missing types

src/widgets/immunizations/immunizations-detailed-summary.component.tsx
```diff
@@ -0,0 +1,167 @@
...
+          isSameImmunization
+        );
+        immunizationConcept.sequences = matchingSequenceDef?.sequences;
+        return immunizationConcept;
+      });
+    };
+  }
+
+  function findExistingDoses(existingImmunizationsForPatient) {
+    return immunizationFromConfig => {
```
FlorianRappl: Missing types

src/widgets/immunizations/immunizations-detailed-summary.component.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+export default function ImmunizationsDetailedSummary(
+  props: ImmunizationsDetailedSummaryProps
+) {
+  const [allImmunizations, setAllImmunizations] = useState(null);
+  const [isLoadingPatient, patient, patientUuid] = useCurrentPatient();
+  const { t } = useTranslation();
+
+  function mapImmunizationSequences(configuredSequences) {
+    return immunizationsConceptSet => {
```
FlorianRappl: Missing types (also the function parameter is not typed)

src/widgets/immunizations/immunizations-detailed-summary.component.tsx
FlorianRappl: Already applies to `existingImmunizationsForPatient`

FlorianRappl (changes requested):
I think we are on the right path here. Thanks for the oustanding work :beers:!

src/widgets/immunizations/immunization-mapper.ts
mddubey: It's actually the Immunizations for a specific patient. I will try to put a Type around this. Probably `FHIRImmunizationBundle` because that's what it is, A [bundle](https://www.hl7.org/fhir/bundle.html) of all immunizations for patient.

src/widgets/immunizations/immunization-mapper.ts
mddubey: It is the FHIR immunization resource. I should be able to resue the one which is sent to server.

src/widgets/immunizations/immunization-mapper.ts
mddubey: Generally coding is an `Array<Code>`. However the idea is to allow only one entry in the code for now, so we have fixed it to be of size one. Let me know if that makes sense.

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: Let me give it a try to better type it. However the code we have referred `ConditionForm` for a lot of initial things. We will see if it could be better represented.

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: Sure, that makes sense! I will change this bit.

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: Ohh, In that case how should this be handled? Are there any existing examples?

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: Atleast not right now. Right now it is coming as module config which is configured as a JSON. We had decided to use the JSON configuration as a label than hardcoding it so it could be translated.

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: Ohh sure! I would add the translations for the confirm box.

src/widgets/immunizations/vaccination-row.component.tsx
mddubey: I think CSS transform makes more sense! We would change this.

src/widgets/immunizations/vaccination-row.component.tsx
mddubey: Just confirming! I should provide `Single Dose on ${recentDose}` as the translated value(for english). And then replace the `{recentDose}` here.

src/widgets/immunizations/vaccination-row.component.tsx
mddubey: Sure!

src/widgets/immunizations/vaccination-row.component.tsx
mddubey: Would provide a type for `patientImmuninization`.

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: I think I get it after reading the comment on the recent vaccine part.

src/widgets/immunizations/immunization-mapper.ts
FlorianRappl: So if its really *always* one entry (i.e., an one-element tuple) then its fine, otherwise I would recommend to use `Array<Code>`.

src/widgets/immunizations/vaccination-row.component.tsx
FlorianRappl: You can use variables in your locale string, e.g.,

```
"foo": "this is my translated string for you, {{name}}",
```

which can then be used like

```
t("foo", { name: "Padawan" })
```

Hope that helps!

mddubey: Thanks! I have implemented it. Didn't know there was a default capability to use formats in the react i18n. Just one confirmation, To get it working I had to use `{name}` instead of `{{name}}`. Single curly brackets.

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: > Does this work? Usually the match params are always strings.

If I understand it clearly, you are referring to the part `props.match.params[0];`. This was a miss on our side. We were sending params as an array with one item. We have fixed it.

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: Regarding the something else instead of `confirm` box if there is an existing approach, we could adopt to that. We checked with few existing forms and they seems to be having `confirm` box only.

src/widgets/immunizations/immunizations-detailed-summary.component.tsx
```diff
@@ -0,0 +1,185 @@
...
+        </table>
+      </SummaryCard>
+    );
+  }
+
+  function displayNoImmunizations() {
+    return (
+      <SummaryCard
+        name={t("Immunizations", "Immunizations")}
+        styles={{
```
FlorianRappl, line 153: Would it make sense to replace the inline style with an import CSS module? Thanks!

src/widgets/immunizations/immunizations-detailed-summary.component.tsx
```diff
@@ -0,0 +1,185 @@
...
+  function displayImmunizations() {
+    return (
+      <SummaryCard
+        name={t("Immunizations", "Immunizations")}
+        styles={{ width: "100%" }}
+      >
+        <table className={`omrs-type-body-regular ${styles.immunizationTable}`}>
+          <thead>
+            <tr>
+              <td>{t("vaccine", "VACCINE")}</td>
```
FlorianRappl: Just for consistency - should we replace these uppercase fallback strings to "normal" ones, i.e., like the others which are normal cased (the casing is done by CSS now, right)?

src/widgets/immunizations/immunizations-form.component.tsx
```diff
@@ -0,0 +1,342 @@
...
+  ] = useCurrentPatient();
+  const formRef = useRef<HTMLFormElement>(null);
+  const { t } = useTranslation();
+  const history = useHistory();
+  const today = new Date().toISOString().split("T")[0];
+  const currentUser = useSessionUser();
+
+  useEffect(() => {
+    if (vaccinationDate) {
+      setEnableCreateButtons(true);
```
FlorianRappl: Usually React is reactive - so you would compute that for your rendering and not after changing, i.e., to ensure it will always stay in sync and reflect what you have. Right now a side-effect is used to trigger another (immediate) state change.

src/widgets/immunizations/immunizations-form.component.tsx
```diff
@@ -0,0 +1,342 @@
...
+    } else {
+      setEnableCreateButtons(false);
+    }
+  }, [vaccinationDate]);
+
+  useEffect(() => {
+    if (viewEditForm && formChanged) {
```
FlorianRappl: This is the same as above - just compute it and avoid unnecessary state changes.

src/widgets/immunizations/immunizations-form.component.tsx
```diff
@@ -0,0 +1,342 @@
...
+        vaccineUuid,
+        manufacturer,
+        expirationDate,
+        vaccinationDate,
+        lotNumber,
+        sequences,
+        currentDose
+      }: ImmunizationFormData = props.match.params;
+
+      setImmunizationObsUuid(immunizationObsUuid);
```
FlorianRappl: All these are re-rendering. Sure we don't want a joint state for them? If you set multiple state entries in the same synchronous code its an indicator that the states are joined and should be treated as one and not independently. In these scenarios, e.g., `useReducer` shines.

src/widgets/immunizations/immunizations-form.component.tsx
```diff
@@ -0,0 +1,342 @@
...
+        immunization,
+        currentVisitUuid,
+        currentLocationUuid,
+        currentProviderUuid
+      ),
+      patientUuid,
+      immunizationObsUuid,
+      abortController
+    ).then(response => {
+      response.status == 200 && navigate();
```
FlorianRappl: Triple equals `response.status === 200`

src/widgets/immunizations/immunizations-form.component.tsx
```diff
@@ -0,0 +1,342 @@
...
+  };
+
+  function navigate() {
+    history.push(`/patient/${patientUuid}/chart/immunizations`);
+    props.closeComponent();
+  }
+
+  const onDoseSelect = event => {
+    const currentDose =
+      immunizationSequences.find(s => s.sequenceNumber == event.target.value) ||
```
FlorianRappl: Triple equals

FlorianRappl (commented):
I think we are quite close here - awesome job! :beers:

src/widgets/immunizations/immunizations-detailed-summary.component.tsx
mddubey: Sure!

src/widgets/immunizations/immunizations-detailed-summary.component.tsx
mddubey: Yes! They are now. Would replace these.

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: If I understand correctly, you are saying I should get rid of this `useEffect` here. And probably use this code of `if(vaccinationDate){}` after the `setVaccinationDate` is called.
Is my understanding correct Or I missed the track here?

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: Sure, Thanks!

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: Hmm! We could do that. Though this is just one case (In case of edit when everything need to be used from previously saved data). I would still need to change the individual state for form field `onSelect` and `onChange`. 

If it is one single object(That's what I understand from joint state), I would need to call something like `setFormState({lotNumber:event.target.value}/setFormState({date:event.target.value}` etc every-time. Is this what you are suggesting?

src/widgets/immunizations/immunizations-form.component.tsx
FlorianRappl: Well the `setVaccinationDate` should be irrelevant - I think you can just compute it (e.g., `const createEnabled = vaccinationDate && ...`). So the only thing that matters are the current state(s) - not how (and in what order) they have been taken.

src/widgets/immunizations/immunizations-form.component.tsx
FlorianRappl: Yes in that direction.

So its either like one object that sets its sub-entries via, e.g.,

```ts
const [formState, setFormState] = React.useState({});
const updateSingle = <T extends keyof typeof formState>(name: T, value: (typeof formState)[T]) => setFormState(state => ({
  ...state,
  [name]: value,
}));
```

or alternatively use the `useReducer`, which gives you this flexibility in a more structured way.

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: Interesting! So if I get you right, you are saying  we can move this write this if clause in a different way out of `useEffect`

```
      const createEnabled = vaccinationDate && true; //want to be explicit here as true
      setEnableCreateButtons(createEnabled);
}
```
And it would reflect automatically whenever the vaccination date is selected.

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: @FlorianRappl I have made this change please have a look! I have used the method `updateSingle`.
Thanks

src/widgets/immunizations/immunizations-detailed-summary.component.tsx
mddubey: Done!

src/widgets/immunizations/immunizations-detailed-summary.component.tsx
mddubey: @FlorianRappl 
I am not able to make this change! If I try to pass `classNames` to SummaryCard it is giving error `classNames` is not present on type `SummaryCardProps`.
I see in the whole app, summarycard sty;es are passed inline.

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: I have tried the above mentioned thing but it is throwing the error saying below as soon as I get it out of the `useEffect`:
> Uncaught Invariant Violation: Too many re-renders. React limits the number of renders to prevent an infinite loop.

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: done

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: done

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: After some reading I thinlk I got what you meant here. You mean I should get rid of `enableCreateButtons/enableCreateButtons` etc as the states and use them more as local variables, which would be computed based on other states.
I have implemented it in the above way now.
Thanks

src/widgets/immunizations/immunizations-form.component.tsx
```diff
@@ -50,19 +50,13 @@ export function ImmunizationsForm(props: ImmunizationsFormProps) {
   const currentUser = useSessionUser();
 
   useEffect(() => {
-    if (formState.vaccinationDate) {
-      updateSingle("enableCreateButtons", true);
-    } else {
-      updateSingle("enableCreateButtons", false);
-    }
+    const enableCreate = !!formState.vaccinationDate;
+    updateSingle("enableCreateButtons", enableCreate);
```
FlorianRappl: Why is this still in an effect? It should not be a side-effect. Just compute it once.

Instead of

```
useEffect(() => {
  // ...
}, [...]);
```

use

```ts
const enableCreate = !!formState.vaccinationDate && formState.viewEditForm && formState.formChanged;
```

src/widgets/immunizations/immunizations-detailed-summary.component.tsx
FlorianRappl: I see - but this is rather a limitation of `SummaryCard`.

https://github.com/VasanthakumarTW/openmrs-esm-patient-chart-widgets/blob/Immunization-Widget/src/ui-components/cards/summary-card.component.tsx#L8

Just change the props to accept an optional `className` and pass this prop to the `className` computation:

```jsx
<div style={props.styles} className={`omrs-card ${styles.card} ${props.className || ''}`}>
```

src/widgets/immunizations/immunizations-form.component.tsx
mddubey: I implemented this feedback in the latest commit after I got the understanding right. It has been changed in the latest commit [here](https://github.com/openmrs/openmrs-esm-patient-chart-widgets/pull/118/commits/de577dce3fae84327090b88db7ab8c234ce18945).

src/widgets/immunizations/immunizations-detailed-summary.component.tsx
FlorianRappl: General remark here: This is why libs like emotion or styled components are better than CSS modules. We could do component inheritance here and override the style without having to fiddle with such details (and then potential problems like CSS specifity / declaration overwriting).

mddubey: Thanks @FlorianRappl!
Somehow I had the understanding since these are all components they should by default support `className` property and I don't need to explicitly define a prop for it. 
I thought classname is a react standard itself.

FlorianRappl (approved):
I think this is definitely good enough for a first release - thanks again for the great work!

LGTM :beers:

src/widgets/immunizations/immunizations-detailed-summary.component.tsx
FlorianRappl: No - there are only three standard props `children`, `key`, and `ref`. And the latter only appears when the component supports it - to support it in custom components you'd need to use something like `forwardRef`.

So we'd need to tell React where / how to use `className`.

FlorianRappl: (Do you want to make this change in this PR or a dedicated one - I guess a dedicated one would make sense as it improves the whole patient-chart-widgets lib)

cc @mddubey

mddubey: Hello @FlorianRappl,
Sorry! I have actually already made this change in the [latest commit here](https://github.com/openmrs/openmrs-esm-patient-chart-widgets/pull/118/commits/b7e6c7275fc73f33dfbedbd0d27ff4963374906e). I have also made changes in the immunization components using it.

If you think this should be dedicated then we could revert last commit.

PR #127, opened 2020-08-03
MF-271: Render attachment according to content family

ivange94:


tendomart:
@ivange94  i think you got a few issues at "/home/travis/build/openmrs/openmrs-esm-patient-chart-widgets/src/widgets/attachments/attachment-thumbnail.component.tsx "

you could have missed terminating a couple of statements .Is it so ?

104:28  error  Insert `·`                                                                                                                                                    prettier/prettier
  112:11  error  Replace `⏎········src={props.src}⏎········alt={props.title}⏎········style={props.style}⏎·····` with `·src={props.src}·alt={props.title}·style={props.style}`  prettier/prettier
  118:4   error  Insert `;`                                                                                                                                                    prettier/prettier
  125:28  error  Replace `'_blank'` with `"_blank"`                                                                                                                            prettier/prettier
  130:7   error  Non-interactive elements should not be assigned mouse or keyboard event listeners                                                                             jsx-a11y/no-noninteractive-element-interactions
  137:4   error  Insert `;`                                                                                                                                                    prettier/prettier
  144:28  error  Replace `'_self'` with `"_self"`                                                                                                                              prettier/prettier
  149:7   error  Non-interactive elements should not be assigned mouse or keyboard event listeners                                                                             jsx-a11y/no-noninteractive-element-interactions
  156:4   error  Insert `;`                                                                                                                                                    prettier/prettier
  168:23  error  Replace `'IMAGE'` with `"IMAGE"`                                                                                                                              prettier/prettier
  169:46  error  Insert `;`                                                                                                                                                    prettier/prettier
  170:30  error  Replace `'PDF'` with `"PDF"`                                                                                                                                  prettier/prettier
  171:44  error  Insert `;`                                                                                                                                                    prettier/prettier
  173:46  error  Insert `;`                                                                                                                                                    prettier/prettier

ivange94:
@FlorianRappl have some time to review this?

src/widgets/attachments/attachment-thumbnail.component.tsx
```diff
@@ -101,59 +101,55 @@ export default function AttachmentThumbnail(props: AttachmentThumbnailProps) {
...
-  )
+  );
 }
 
 function OtherThumbnail(props: ImageProps) {
   function handleClick(e: React.SyntheticEvent) {
     e.preventDefault();
     e.stopPropagation();
-    window.open(props.src, '_self');
+    window.open(props.src, "_self");
```
FlorianRappl: Why not a new window?

FlorianRappl:
@ivange94  - yes, looks good so far. I only have one question (above). Many thanks! :beers:

src/widgets/attachments/attachment-thumbnail.component.tsx
ivange94: Is there a better way to do this? I just want to be able to download this file and I went with this solution from searching the internet.

FlorianRappl: But you don't know if what you open actually triggers a download. If its a TXT you'd just open it...

For downloads a better solution would be to dynamically append an `a` to the DOM with the `download` attribute. Clicking it would trigger a download (guaranteed).

Then afterwards just remove the  `a`.

FlorianRappl:
Any update on this one @ivange94 ?

ivange94:
@FlorianRappl sorry I missed this and somehow felt this was already closed. I'll look into this tomorrow morning.

FlorianRappl:
Hi @ivange94 just wanted to check if you are still on this or how we should proceed here. Thanks! :beers:

FlorianRappl:
> Hi @ivange94 just wanted to check if you are still on this or how we should proceed here. Thanks! 🍻

Any status updates @ivange94 ?

src/widgets/attachments/attachment-thumbnail.component.tsx
ivange94: Been a while I worked on this component but the solution you proposed isn't working I guess that was why I didn't go with that approach earlier. And it has to do probably with the library used. I'm still investigating this but might take some time.

ivange94: The library event handlers seem to be overriding every other behavior and even with a 

`<a href={linkToFile} download/>` used, on clicking the file is just opened in full screen mode.

ivange94:
@FlorianRappl I've applied your suggestion and it works fine. All work on this is now complete.

FlorianRappl (approved):
Awesome thanks Larry!

PR #130, opened 2020-08-05
Fixed broken links on patient-summary

donaldkibet:
1. This PR fixes broken links on patient-summary. At the moment `basePath` is passed as an `object` but used as a `string`.

linkFix

### DIFF
--- a/src/widgets/allergies/allergies-overview.component.tsx
+++ b/src/widgets/allergies/allergies-overview.component.tsx
@@ -16,6 +16,7 @@ import {
 } from "./allergy-intolerance.resource";
 import AllergyForm from "./allergy-form.component";
 import styles from "./allergies-overview.css";
+import { widgetBasePath } from "../types";
 
 export default function AllergiesOverview(props: AllergiesOverviewProps) {
   const initialAllergiesBatchCount = 3;
@@ -28,8 +29,11 @@ export default function AllergiesOverview(props: AllergiesOverviewProps) {
   const [allergiesExpanded, setAllergiesExpanded] = useState<boolean>(false);
   const [isLoadingPatient, patient, patientUuid] = useCurrentPatient();
   const chartBasePath = useChartBasePath();
-  const allergiesPath = chartBasePath + "/" + props.basePath;
   const { t } = useTranslation();
+  const {
+    props: { basePath }
+  } = props;
+  const allergiesPath = chartBasePath + "/" + basePath;
 
   useEffect(() => {
     if (!isLoadingPatient && patient) {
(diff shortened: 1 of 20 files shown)


src/widgets/allergies/allergies-overview.component.tsx
```diff
@@ -28,8 +29,11 @@ export default function AllergiesOverview(props: AllergiesOverviewProps) {
   const [allergiesExpanded, setAllergiesExpanded] = useState<boolean>(false);
   const [isLoadingPatient, patient, patientUuid] = useCurrentPatient();
   const chartBasePath = useChartBasePath();
-  const allergiesPath = chartBasePath + "/" + props.basePath;
   const { t } = useTranslation();
+  const {
+    props: { basePath }
+  } = props;
+  const allergiesPath = chartBasePath + "/" + basePath;
```
brandones: Widgets are receiving a prop called `props`??? That seems weird and bad. Perhaps I messed something up in https://github.com/openmrs/openmrs-esm-patient-chart/blob/master/src/view-components/widget/widget.component.tsx . The `widgetConfig.props` should be passed through as the actual widget props.

src/widgets/allergies/allergies-overview.test.tsx
```diff
@@ -39,12 +40,14 @@ describe("<AllergiesOverview />", () => {
...
 
+  const mockWidgetBasePath: widgetBasePath = { basePath: "/" };
+
   it("should display the patient's allergic reactions and their manifestations", async () => {
     mockPerformPatientAllergySearch.mockReturnValue(of(mockPatientAllergies));
 
     render(
       <BrowserRouter>
-        <AllergiesOverview basePath="/" />
+        <AllergiesOverview props={mockWidgetBasePath} />
```
brandones: Yeah, something has gone terribly wrong. Let's fix the underlying problem, which is the mechanism by which props are passed to the widgets.

src/widgets/types.ts
```diff
@@ -247,3 +247,7 @@ type DisplayMetadata = {
   links?: Links;
   uuid?: string;
 };
+
+export interface widgetBasePath {
```
ivange94: I think this is really confusing for a type name. My mind constantly reads this as a variable name(value) when imported. Why not call this `WidgetBasePath` following the other patterns in the file?

src/widgets/allergies/allergies-overview.test.tsx
donaldkibet: I have pushed a fixed where I pass basePath as a prop, check out this PR [patient-chart](https://github.com/openmrs/openmrs-esm-patient-chart/pull/175).

PR #131, opened 2020-08-08
FM-300:Use REST to fetch relationships until FHIR endpoint is added

reagan-meant:
issue: https://issues.openmrs.org/browse/MF-300

### DIFF
--- a/src/widgets/profile/relationships.card.test.tsx
+++ b/src/widgets/profile/relationships.card.test.tsx
@@ -39,6 +39,7 @@ describe("<RelationshipsCard/>", () => {
     expect(wrapper.getByText("\u2014")).not.toBeNull();
   });
 
+  /* TO BE UNCOMMENTED WHEN USING FHIR
   it("renders relationships correctly", async () => {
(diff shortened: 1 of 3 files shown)


src/widgets/profile/relationships.card.test.tsx
```diff
@@ -39,6 +39,7 @@ describe("<RelationshipsCard/>", () => {
     expect(wrapper.getByText("\u2014")).not.toBeNull();
   });
 
+  /* TO BE UNCOMMENTED WHEN USING FHIR
```
HerbertYiga, line 42: @reagan-meant  well done on this.Does this mean this section will always be uncommented when one wants to use FHIR,ie a non developer ?

reagan-meant: @HerbertYiga this is a test i commented hence it doesn't affect a non developer....Unfortunately you missed a background to this ticket....This is a temporally fix...[See here](https://talk.openmrs.org/t/fhir-url-for-relatedperson/29624)

brandones (approved):
Thanks @reagan-meant !

PR #132, opened 2020-09-08
MF-323: Fix allergy widget display property resolution

denniskigen:
The display property for allergies is not working again following a backend tweak referenced here https://issues.openmrs.org/browse/FM2-298 which necessitates a change in the way we resolve the display concept on the frontend.

I've changed the way we access the deeply nested display property from the API response to get it back working.

### DIFF
--- a/src/widgets/allergies/allergy-intolerance.resource.tsx
+++ b/src/widgets/allergies/allergy-intolerance.resource.tsx
@@ -20,18 +20,18 @@ export function performPatientAllergySearch(patientIdentifier: string) {
 }
 
 export function fetchAllergyByUuid(allergyUuid: string) {
-  return openmrsObservableFetch<Allergy>(
+  return openmrsObservableFetch(
     `${fhirBaseUrl}/AllergyIntolerance/${allergyUuid}`
   ).pipe(
     map(({ data }) => data),
-    map(data => mapAllergyProperties(data))
+    map((data: FHIRAllergy) => mapAllergyProperties(data))
   );
 }
 
-function mapAllergyProperties(allergy) {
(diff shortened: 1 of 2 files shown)


src/widgets/allergies/allergy-intolerance.resource.tsx
```diff
@@ -31,13 +31,13 @@ export function fetchAllergyByUuid(allergyUuid: string) {
 function mapAllergyProperties(allergy) {
```
FlorianRappl: Missing type (sorry I know this is not your change, but otherwise its hard to reason about the rest)

src/widgets/allergies/allergy-intolerance.resource.tsx
```diff
@@ -31,13 +31,13 @@ export function fetchAllergyByUuid(allergyUuid: string) {
...
   let manifestations: Array<string> = [];
   allergy?.reaction[0]?.manifestation?.map(coding =>
-    manifestations.push(coding.coding[1].display)
+    manifestations.push(coding.coding[0].display)
```
FlorianRappl: Sure `[0]?.display` isn't more suitable?

src/widgets/allergies/allergy-intolerance.resource.tsx
denniskigen: Updated.

src/widgets/allergies/allergy-intolerance.resource.tsx
denniskigen: It is indeed.

FlorianRappl (approved):
LGTM :beers:!

PR #138, opened 2020-09-23
MF-219: Add visits widget translations

denniskigen:

- Adds translation keys to the components that make up the visits widget as well as minimal improvements to styling and tests.
- Separates out the `Start date` and `Start time` input labels so they operate independent of each other. A positive corollary of this is that we no longer need to use data-testids to query DOM elements when testing which is not ideal anyway. This has potentially unearthed a subtle bug with the way the Visit stop time is determined when editing a visit. I will investigate further and potentially fix this in a subsequent PR.

Screenshot 2020-09-23 at 20 13 56
Screenshot 2020-09-23 at 20 06 44
Screenshot 2020-09-23 at 20 07 21

### DIFF
--- a/__mocks__/visits.mock.ts
+++ b/__mocks__/visits.mock.ts
@@ -7,7 +7,7 @@ export const mockVisitTypes = [
   {
     uuid: "some-uuid2",
     name: "HIV Return Visit",
-    display: "Outpatient Visit"
+    display: "HIV Return Visit"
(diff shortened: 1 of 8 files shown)


FlorianRappl (approved):
LGTM!

FlorianRappl (approved):
Thanks for the work! 🍻

PR #139, opened 2020-09-30
[MF-284] Align 'Edit' button to work correctly with esm-config's navigator used in the patient registration component

samuelmale:
**Parent PR**
> https://github.com/openmrs/openmrs-esm-patient-registration/pull/51

**Ticket**
> https://issues.openmrs.org/browse/MF-284

### DIFF
--- a/src/widgets/banner/patient-banner.component.tsx
+++ b/src/widgets/banner/patient-banner.component.tsx
@@ -67,7 +67,8 @@ export default function PatientBanner(props: PatientBannerProps) {
               to={
                 "/patient-registration/patient/" +
                 patient.id +
-                "?afterUrl=${openmrsSpaBase}" +
+                "?afterUrl=" +
+                window.getOpenmrsSpaBase() +
                 pathname
               }
               className="omrs-link omrs-text-neutral"
--- a/src/widgets/banner/patient-banner.component.tsx
+++ b/src/widgets/banner/patient-banner.component.tsx
@@ -67,7 +67,7 @@ export default function PatientBanner(props: PatientBannerProps) {
               to={
                 "/patient-registration/patient/" +
                 patient.id +
-                "?afterUrl=" +
+                "?afterUrl=${openmrsSpaBase}" +
                 pathname
               }
               className="omrs-link omrs-text-neutral"
--- a/src/widgets/banner/patient-banner.component.tsx
+++ b/src/widgets/banner/patient-banner.component.tsx
@@ -68,7 +68,7 @@ export default function PatientBanner(props: PatientBannerProps) {
                 "/patient-registration/patient/" +
                 patient.id +
                 "?afterUrl=" +
-                window.getOpenmrsSpaBase() +
+                (window as any).getOpenmrsSpaBase().replace(/\/$/, "") +
                 pathname
               }
               className="omrs-link omrs-text-neutral"


brandones:
EDIT: whoops, just saw that this is a different repo from https://github.com/openmrs/openmrs-esm-patient-registration/pull/51, sorry! Lemme look.

brandones:
I think it should be the responsibility of whoever *sets* the `afterUrl` to include the correct URL base. Does that make sense?

brandones:
Wait, sorry, it's taken me a minute to figure out what's going on here. Here, you are in fact setting `afterUrl`, so my previous comment doesn't make sense. However, we should never be passing magic template stuff in URL parameters like this. Does the existing code not work correctly? It seems like this change shouldn't be necessary at all.

samuelmale:
@brandones The existing code works correctly as is. But as soon as we have https://github.com/openmrs/openmrs-esm-patient-registration/pull/51 merged in, it will break. This is because the `afterUrl` won't contain the `openmrsSpaBase` yet the [navigate({ to: string })](https://github.com/openmrs/openmrs-esm-module-config/blob/2c014c2f954ad700f9be0eec1f1ccc7d8d8d5aee/src/navigation/navigate.ts#L24-L33) expects the URL to have the base or rather something close to an absolute path hence the redirect will be broken. 

In short, in the parent PR I replaced `history.push(..)` from react-router with `navigate(..)` from esm-config.

brandones:
`navigate` can accept any kind of path. Passing it a root-relative path will behave as expected. `afterUrl` should not accept template parameters.

EDIT: Or rather, we should treat `afterUrl` as if it does not accept template parameters.

src/widgets/banner/patient-banner.component.tsx
```diff
@@ -67,7 +67,7 @@ export default function PatientBanner(props: PatientBannerProps) {
               to={
                 "/patient-registration/patient/" +
                 patient.id +
-                "?afterUrl=" +
+                "?afterUrl=${openmrsSpaBase}" +
```
brandones:
```suggestion
                "?afterUrl=" +
                window.getOpenmrsSpaBase() +
```

samuelmale:
@brandones applied commit suggestion. Thanks!

brandones:
No problem. Looks like there's a Typescript failure. Try `(window as any).getOpenmrsSpaBase()`

samuelmale:
> No problem. Looks like there's a Typescript failure. Try `(window as any).getOpenmrsSpaBase()`

Oya, but I have some early findings:

- `window.getOpenmrsSpaBase()` returns `/openmrs/spa/`
- `pathname` from the `useLocation()` hook will be `/patient/54c77c2c-a53f-41ca-a167-705ae0141b9f/chart/summary`

Appending the base to the pathname results to something like: `/openmrs/spa//patient/54c77c2c-a53f-41ca-a167-705ae0141b9f/chart/summary`

There are trailing slashes: `spa//patient`

samuelmale:
@brandones I think this is why this exists: https://github.com/openmrs/openmrs-esm-module-config/blob/2c014c2f954ad700f9be0eec1f1ccc7d8d8d5aee/src/navigation/navigate.ts#L25

samuelmale:
@brandones Fixed.

brandones (approved):
Awesome, thanks Samuel! Yeah, getting the current root path could be a little easier than it presently is. Maybe something we'll add to esm-core at some point.

PR #141, opened 2020-10-12
New App Shell Tooling

FlorianRappl:
- Consolidated the package.json
- Introduced the `openmrs` tooling
- Updated dependencies


brandones (approved):
Asked a question on Slack about how we're using `dependencies`, but LGTM!

PR #147, opened 2020-10-26
Fix failing ConditionRecord test

denniskigen:
Fixes test that was failing due to a missing property. This fixes CI builds as well.


brandones (approved):
I don't know anything about the conditions data model. But assuming it's allowed that it can be missing `lastUpdated` and `lastUpdatedLocation`, this looks good.

denniskigen:
@brandones the Conditions resource does not provide this metadata in its current state. The code assumed that the `lastUpdated` property existed and tried to call `dayjs(...).format()` on it which ends up returning a string for the current date. This worked on the day I pushed the code which is why it slipped through.

PR #149, opened 2020-11-11
MF-374: Create reusable DataTable and EmptyState components

denniskigen:
https://issues.openmrs.org/projects/MF/issues/MF-374

This PR is the first in a line of changes aimed at introducing carbon support to the patient chart summary. It:

- Adds SASS support to the Webpack config - a pre-requisite for working with Carbon styles.
- Creates a reusable DataTable component which we'll use to display the data in our widgets.
- Modifies the existing EmptyState component so that it can be used to display both empty states and error management states.
- Removes references from tests that asserted for the presence of an Add button in the EmptyState component card header. We're removing this in favor of a link that the user can click to launch a workspace tab per the new designs. This is the reason for the large diff noise (apologies!).

Screenshot 2020-11-11 at 16 46 50


denniskigen:
**Known issues:** 
- Slow SASS compile times.
- SVG color is being overridden by something; not sure what, yet.

package.json
```diff
@@ -43,6 +43,7 @@
     "@fortawesome/fontawesome-svg-core": "^1.2.29",
     "@fortawesome/free-solid-svg-icons": "^5.13.1",
     "@fortawesome/react-fontawesome": "^0.1.11",
+    "@openmrs/react-root-decorator": "^3.3.1",
```
FlorianRappl: In the new version this was replaced IIRC by the context module(`@openmrs/esm-context`).

webpack.config.js
```diff
@@ -55,6 +55,17 @@ module.exports = env => ({
             }
           }
         ]
+      },
+      {
+        test: /\.s[ac]ss$/i,
+        use: [
+          // Creates `style` nodes from JS strings
+          "style-loader",
```
FlorianRappl, line 63: Should we use this or an external CSS file?

FlorianRappl (commented):
Awesome work!

webpack.config.js
denniskigen: I don't know. Is the common convention to use a file instead?

FlorianRappl: What I meant was:

With the `style-loader` we put the CSS into a `<style>` tag. The CSS itself is all transported via JS. If we'd use the CSS extraction plugin we'd get a dedicated file and the loader would need to create a `<link>` instead.

Normally, I'd prefer the latter as it gives better performance in a number of scenarios (caching, discoverability, ...).

FlorianRappl: (and yes, I'd say the standard is to use CSS in a `<link>` / external file instead of `<style>`; similar to JS which is usually used in dedicated files and not inline)

package.json
brandones: +1, if you could replace this, that would be great

webpack.config.js
brandones: @denniskigen Do you know how to move forward with this? @FlorianRappl makes a sound argument.

webpack.config.js
```diff
@@ -55,6 +55,17 @@ module.exports = env => ({
...
+          // Translates CSS into CommonJS
+          "css-loader",
+          // Compiles Sass to CSS
```
brandones, line 66: Also, bless you for adding these comments, I can never for the life of me remember what these loaders do

webpack.config.js
```diff
@@ -55,6 +55,17 @@ module.exports = env => ({
             }
           }
         ]
+      },
+      {
+        test: /\.s[ac]ss$/i,
+        use: [
+          // Creates `style` nodes from JS strings
+          "style-loader",
+          // Translates CSS into CommonJS
```
brandones, line 64: Also, bless you for adding these comments, I can never for the life of me remember what these loaders do

src/ui-components/datatable/datatable.component.tsx
```diff
@@ -0,0 +1,105 @@
...
+  TableHead,
+  TableHeader,
+  TableRow,
+  TableToolbar,
+  TableToolbarContent
+} from "carbon-components-react";
+import { Add16 } from "@carbon/icons-react";
+import "./datatable.scss";
+
+const WidgetDataTable = ({
```
brandones, line 21: How or where are these components actually used?

src/ui-components/datatable/datatable.component.tsx
```diff
@@ -0,0 +1,105 @@
...
+  title,
+  rows,
+  headers,
+  linkTo,
```
brandones, line 25: Link to what? Maybe this should be `seeAllLinkTarget` or something?

brandones: Or `detailViewLink`, something like that

src/ui-components/datatable/datatable.component.tsx
```diff
@@ -0,0 +1,105 @@
...
+      }) => (
+        <TableContainer title={title}>
+          <TableToolbar>
+            <TableToolbarContent>
+              {addComponent && (
+                <Button
+                  kind="ghost"
+                  renderIcon={Add16}
+                  iconDescription="Add"
+                  onClick={() => showComponent(addComponent, name)}
```
brandones: Where does `name` come from?

src/ui-components/datatable/datatable.component.tsx
```diff
@@ -0,0 +1,105 @@
...
+                    {header.header}
+                  </TableHeader>
+                ))}
+              </TableRow>
+            </TableHead>
+            <TableBody>
+              {rows.map(row => (
+                <TableRow key={row.id} {...getRowProps({ row })}>
+                  {row.cells.map(cell => (
+                    <TableCell key={cell.id}>{cell.value}</TableCell>
```
brandones, line 75: Hopefully these `row.id` and `cell.id` are very unique?

src/ui-components/empty-state/empty-state.component.tsx
```diff
@@ -1,40 +1,97 @@
...
+          <Table>
+            <TableBody>
+              <TableRow>
+                <TableCell>
+                  {props.hasError ? <ErrorManagementView /> : <EmptyDataView />}
+                </TableCell>
+              </TableRow>
+            </TableBody>
+          </Table>
+        </TableContainer>
```
brandones, line 78: What's the point of all that wrapping around the error / empty data components?

brandones:
Really good, generally. One thing looks like a bug. A couple outstanding notes from @FlorianRappl . I'd really like to see the piece of code that uses `DataTableWidget` so I can understand what `showComponent` will generally look like.

brandones (changes requested):
See notes above

src/ui-components/datatable/datatable.component.tsx
denniskigen: See this [commit](https://github.com/openmrs/openmrs-esm-patient-chart-widgets/commit/f4cdda1b59df00588f2b59322035eea991916944) where I refactor the existing widgets to reuse these new components.

webpack.config.js
denniskigen: I don't. Would you mind expounding on exactly what would need to be done here @FlorianRappl?

src/ui-components/datatable/datatable.component.tsx
denniskigen: Sure, `detailViewLink` sounds clearer.

src/ui-components/datatable/datatable.component.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { Add16 } from "@carbon/icons-react";
+import "./datatable.scss";
+
+const WidgetDataTable = ({
+  title,
+  rows,
+  headers,
+  linkTo,
+  addComponent,
+  showComponent
```
brandones: Looking at the API used in [the example](https://github.com/openmrs/openmrs-esm-patient-chart-widgets/commit/f4cdda1b59df00588f2b59322035eea991916944#diff-7c5e3c1f85a66a10957b2ee1a79ab33f87d1bdc60dc611e5be6583c42244628aR86-R91), why did you add `addComponent` to the API, and make `showComponent` take two arguments? Seems simpler to have it be a nullary function and no separate `addComponent` parameter.

brandones: Also I'd suggest `showAddComponent` as a name. `showComponent` leaves the question "what component?"

Unless it's intended to be used more generally than just for the `add` component?

src/widgets/allergies/allergies-detailed-summary.component.tsx
```diff
@@ -148,7 +148,8 @@ export default function AllergiesDetailedSummary(
       ) : (
         <EmptyState
           name="Allergies"
-          showComponent={() =>
+          displayText={t("allergy intolerances", "allergy intolerances")}
```
brandones, line 151:
```suggestion
          displayText={t("allergyIntolerances", "allergy intolerances")}
```

src/ui-components/datatable/datatable.component.tsx
brandones: Ping

brandones (approved):
One typo and one remaining old suggestion. But LGTM! Thanks for this!

brandones:
Oh wait, probably best to wait for @FlorianRappl to respond to the question above and get his approval as well.

webpack.config.js
FlorianRappl: Yeah we would need to change the `style-loader` and use, e.g., the `mini-css-extract-plugin` instead. This way, we get a `.css` file instead of a `<style>` tag. The plugin comes with a loader, which should place the `<link>` instead.

FlorianRappl (commented):
In general looks good to me.

I would, however, put shared components into esm-styleguide. If we want to make these components truly re-usable, then this is the way to go I'd argue.

denniskigen:
@FlorianRappl I agree. Thank you and Brandon for the valuable feedback. I'm closing this PR in lieu of the discussions we've had offline and especially informed by the call we had earlier. I'll make sure to reintroduce the useful parts of this work in future.

PR #151, opened 2020-11-22
MF-382 Update vitals widgets to match latest designs (Overview widgets)

donaldkibet:
### What this PR does?

It starts the vitals widget carbon designs integration


src/extensions/spa-vitals-extension.tsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import ReactDOM from "react-dom";
+import singleSpaReact from "single-spa-react";
+import ExtensionRoot from "./vitals-extesnion";
```
jonathandick: the above has a misspelling

src/extensions/spa-vitals-overview-extension.tsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
```
jonathandick: @donaldkibet , I don't think we need this overview-extension. In fact, we have a few layers here which don't exactly make sense to me:
spa-vitals-overview-extension -> vitals-overview-chart-extension -> vitals-chart-overview

I think this could/should be simplified to :
vitals-extension -> vitals.component

Is there a reason to have the above structure instead? 

@brandones, @FlorianRappl ,  what pattern do you recommend for extensions in terms of file organization? My sense is that .extension should be the wrapping of a component in single-spa. This should import the Root component. Or should it all just be in one file? This is the time to set a best practice for this.

FlorianRappl: Each extension has its own root component. I'd argue we should make one folder per extension. Every folder comes with the single-spa root declaration, the root module, and everything else required to make it work. Naturally, I'd take every widget directory in this project as an extension.

FlorianRappl (approved):
LGTM - thanks!

PR #153, opened 2020-11-29
Improvements to relationships

denniskigen:
This PR improves the relationships feature by:

- Adding type annotations.
- Modifying the request so that it returns the full relationships response (hat tip @jonathandick).
- Adding logic that allows the Relationship component to distinguish between `AisToB` relations and `BisToA` relations and display them accordingly.
- Adding styling that ensures the UI is similar to that provided in the designs.

relationships

### DIFF
--- a/src/widgets/contact-details/relationships.resource.tsx
+++ b/src/widgets/contact-details/relationships.resource.tsx
@@ -2,11 +2,27 @@ import { openmrsFetch } from "@openmrs/esm-api";
 
 export function fetchPatientRelationships(patientIdentifier: string) {
   return openmrsFetch<{ results: Relationship[] }>(
-    `/ws/rest/v1/relationship?person=${patientIdentifier}`
+    `/ws/rest/v1/relationship?v=full&person=${patientIdentifier}`
   );
 }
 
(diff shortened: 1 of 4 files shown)


src/widgets/contact-details/relationships.resource.tsx
```diff
@@ -2,11 +2,27 @@ import { openmrsFetch } from "@openmrs/esm-api";
 
 export function fetchPatientRelationships(patientIdentifier: string) {
   return openmrsFetch<{ results: Relationship[] }>(
-    `/ws/rest/v1/relationship?person=${patientIdentifier}`
+    `/ws/rest/v1/relationship?v=full&person=${patientIdentifier}`
```
jonathandick, line 5: @denniskigen this will likely return a lot of unneeded data. We can merge for now but I would suggest using a more custom request than full.

jonathandick (approved):

src/widgets/contact-details/relationships.resource.tsx
denniskigen: @jonathandick fixed in #155

PR #154, opened 2020-11-29
Refactor tests to reuse mocks

denniskigen:
Tests should reuse mocks for `useCurrentPatient` and the mock `Patient` object instead of having individual spec files redefine them anew.


FlorianRappl (approved):
Superb - thanks :beers:!

PR #155, opened 2020-11-30
Modify relationships request query params to use custom representation

denniskigen:
Switch from `full` to custom representation in request URL query params.


FlorianRappl (approved):
Alright, for me this is all fine - thanks! :beers:

PR #159, opened 2020-12-01
Contact details button should toggle caret types

denniskigen:
Hat tip to Ciaran for spotting this - clicking the button should toggle between the caret pointing up and down as well.

contact-details-button-caret


brandones (approved):
Thanks!

PR #162, opened 2020-12-02
[MF-295] Patient Chart Widgets should be extensions

ivange94:


.travis.yml
```diff
@@ -30,7 +30,7 @@ after_deploy:
   - >
     if [[ "$TRAVIS_BRANCH" == "master" ]]; then
       echo "Updating import map to point to new version of @openmrs/esm-patient-chart-widgets"
-      statuscode=$(curl --output /dev/null --write-out %{http_code} -u $DEPLOYER_USERNAME:$DEPLOYER_PASSWORD -d '{ "service":"@openmrs/esm-patient-chart-widgets","url":"https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-patient-chart-widgets/'$TRAVIS_COMMIT'/openmrs-esm-patient-chart-widgets.js" }' -X PATCH $DEPLOYER_HOST/services\?env=prod -H "Accept:application/json" -H "Content-Type:application/json")
+      statuscode=$(curl --output /dev/null --write-out %{http_code} -u $DEPLOYER_USERNAME:$DEPLOYER_PASSWORD -d '{ "service":"@openmrs/esm-patient-chart-widgets-app","url":"https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-patient-chart-widgets/'$TRAVIS_COMMIT'/openmrs-esm-patient-chart-widgets.js" }' -X PATCH $DEPLOYER_HOST/services\?env=prod -H "Accept:application/json" -H "Content-Type:application/json")
```
FlorianRappl: Why? Can we leave it without the `-app`? Its just a collection of extensions.

src/index.ts
```diff
@@ -48,6 +48,39 @@ function setupOpenMRS() {
...
         )
+      },
+      {
+        id: "attachments-overview-widget",
+        slot: "attachments-overview-tab",
+        load: getAsyncLifecycle(
+          () => import("./widgets/attachments/attachments-overview.component"),
+          {
+            featureName: "attachments",
+            moduleName: "@openmrs/esm-patient-chart-widgets"
```
FlorianRappl: The `moduleName` can be placed in a local variable `const moduleName = ...` and could be reused.

src/widgets/attachments/attachments-overview.extension.ts
```diff
@@ -0,0 +1,12 @@
+import React from "react";
```
FlorianRappl: Not needed, or? Replaced by `getAsyncLifecycle `

src/widgets/immunizations/immunizations.component.tsx
```diff
@@ -1,21 +1,20 @@
...
-
   return (
-    <Switch>
-      <Route exact path={match.path}>
-        <ImmunizationsDetailedSummary
-          immunizationsConfig={props.config.immunizationsConfig}
-        />
-      </Route>
-    </Switch>
+    <BrowserRouter basename={window["getOpenmrsSpaBase"]()}>
```
FlorianRappl, line 9: Its a single route definition. Do we need this (the question applies to all the widgets using this pattern)?

Is this about using `useRouteMatch` / using the `patientUuid` parameter in the path / using some `<Link>` inside?

Personally, I think the `patientUuid` should be transported via props to the extension slot (defined in patient-chart).

.travis.yml
ivange94: This is a mistake. I'd made this change after my conversation with you on slack but improvements were later made that made this change unnecessary. I just forgot to undo it. Will fix now.

src/widgets/allergies/allergies.component.tsx
```diff
@@ -1,20 +1,20 @@
...
-      <Route exact path={`${path}`}>
-        <AllergyDetailedSummary />
-      </Route>
-      <Route exact path={`${path}/:allergyUuid`}>
-        <AllergyRecord />
-      </Route>
-    </Switch>
+    <BrowserRouter basename={window["getOpenmrsSpaBase"]()}>
+      <Switch>
+        <Route exact path="/patient/:patientUuid/chart/allergies">
```
FlorianRappl, line 10: No we don't know that (URL). This is a URL related to patient-chart, but the extension may be used in other components (page), too. The same applies to all other extension components in this PR.

Ideally, the extension slot is supplied a patient UUID and a basePath. So you can derive your routes here via `${basePath}/allergies/...`.

FlorianRappl (commented):
I think the direction is right, now I need to change patient-chart to supply these props.

It would be good though that you already expect (and check) these props. Otherwise the other side of the tunnel is finished, but you did not start digging yet.

ivange94:
@FlorianRappl @brandones this piece of code 

```typescript
jest.mock("react-router-dom", () => {
  const originalModule = jest.requireActual("react-router-dom");

  return {
    __esModule: true,
    ...originalModule,
    BrowserRouter: jest
      .fn()
      .mockImplementation(({ children }) => <div>{children}</div>)
  };
});

window["getOpenmrsSpaBase"] = jest.fn().mockImplementation(() => "/");
```
appears in a lot of places. So I thought of reducing that by add a `react-router-dom.tsx` to `__mocks__` so all test will just use this mock. But this made the broken tests move from 11 to 48. Following this pattern would have required me to modify code that has nothing to do with this ticket. 

Another approach I thought of was to add an utility method and call this method in tests that need this functionality. This worked for all but one test `appointments.component.test.tsx`. Not yet sure why that particular test stands out.

Do you have any suggestions for dealing with this?

src/widgets/allergies/allergies.component.tsx
ivange94: The `basePath` prop returned undefined so I just left it out. Once They're available, I'll refactor.

src/widgets/allergies/allergies-overview.component.tsx
```diff
@@ -27,7 +28,7 @@ export default function AllergiesOverview(props: AllergiesOverviewProps) {
   );
   const [allergiesExpanded, setAllergiesExpanded] = useState<boolean>(false);
   const [isLoadingPatient, patient, patientUuid] = useCurrentPatient();
-  const chartBasePath = useChartBasePath();
+  const chartBasePath = "/patient/:patientUuid/chart";
```
FlorianRappl, line 31: Should not be known / hardcoded here.

src/widgets/allergies/allergies-overview.component.tsx
```diff
@@ -58,7 +59,7 @@ export default function AllergiesOverview(props: AllergiesOverviewProps) {
   };
 
   return (
-    <>
+    <BrowserRouter basename={window["getOpenmrsSpaBase"]()}>
```
FlorianRappl, line 62: Should not be needed here. Instead, we should use the provided `basePath` prop.

src/widgets/allergies/allergies.component.tsx
FlorianRappl: Should be available.

src/widgets/appointments/appointments-overview.component.tsx
```diff
@@ -18,7 +18,7 @@ export default function AppointmentsOverview(props: AppointmentOverviewProps) {
   const [patientAppointments, setPatientAppointments] = useState([]);
   const [isLoadingPatient, , patientUuid] = useCurrentPatient();
   const startDate = dayjs().format();
-  const chartBasePath = useChartBasePath();
+  const chartBasePath = "/patient/:patientUuid/chart";
```
FlorianRappl, line 21: Same as above. Should not be needed / hardcoded. Instead use `basePath`.

src/widgets/appointments/appointments.component.tsx
```diff
@@ -1,19 +1,22 @@
...
-      <Route exact path={match.path}>
-        <AppointmentsDetailedSummary />
-      </Route>
-      <Route exact path={`${match.path}/:appointmentUuid`}>
-        <AppointmentRecord />
-      </Route>
-    </Switch>
+    <BrowserRouter basename={window["getOpenmrsSpaBase"]()}>
+      <Switch>
+        <Route exact path="/patient/:patientUuid/chart/appointments">
```
FlorianRappl, line 10: This may not be the right path. Always pick the path from `basePath`.

src/widgets/immunizations/immunizations-overview.component.tsx
```diff
@@ -46,44 +46,46 @@ export default function ImmunizationsOverview(
...
+              }}
+              value={t("recentVaccination", "Recent Vaccination")}
+              valueStyles={{
+                color: "var(--omrs-color-ink-medium-contrast)",
+                fontFamily: "Work Sans"
+              }}
+            />
+          </SummaryCardRowContent>
+        </SummaryCardRow>
+        {patientImmunizations &&
```
FlorianRappl, line 71: Instead of `foo && foo....` you can also do `foo?...` like

```ts
patientImmunizations?.map(immunization => { ... })
```

FlorianRappl (commented):
I think we are on the right track here. Now we need to resolve the conflicts and then use the provided props.

cc @ivange94

gracepotma:
@FlorianRappl to review again per our discussion today so we can merge this :)

FlorianRappl:
Is this still active?

cc @ivange94

FlorianRappl:
Ping @ivange94

jonathandick:
My sense is this PR is no longer relevant. Can we close this @ivange94 ?

PR #163, opened 2020-12-07
Tighten up margins of elements in the patient banner

denniskigen:
Following on from this feedback from Ciaran: 

Screenshot 2020-12-07 at 09 40 38

Before:

Screenshot 2020-12-07 at 09 39 55

After:

copy without baseline

Is the use of negative margins here egregious?


FlorianRappl (approved):
Let's give it a shot.

The negative margin would potentially not be required depending on the parent styling. However, since we cannot control it here - let's leave it as is.

PR #165, opened 2020-12-07
MF:383 Record Vitals Page

donaldkibet:
### What does this PR do? 

- Carbonise Vitals and Biometric form 
- Remove old components not needed according to the new designs

### Screenshoot
Screenshot 2020-12-08 at 19 56 21


jonathandick:
This looks good to me. @FlorianRappl ?

__mocks__/chart-widgets-config.mock.ts
```diff
@@ -10,6 +10,17 @@ export const ConfigMock = {
...
+    oxygenSaturationUnit: "%",
+    pulseUnit: "bpm",
+    respiratoryRateUnit: "/min",
+    temperatureUnit: "°C",
+    midUpperArmCircumferenceUnit: "cm"
+  },
+  biometrics: {
+    bmiUnit: "kg / m²",
+    heightUnit: "cm",
+    weightUnit: "kg"
```
brandones: All these units except the bmiUnit should come from the corresponding concept metadata.

src/config-schema.ts
```diff
@@ -18,15 +18,15 @@ const schema = {
       _type: Type.ConceptUuid,
       _default: "5086AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
     },
-    pulseUuid: {
+    heartRateUuid: {
```
brandones: Where did these names come from? A while back I standardized them to the CIEL Fully Specified Names. e.g. https://openconceptlab.org/orgs/CIEL/sources/CIEL/concepts/5087/

src/config-schema.ts
```diff
@@ -38,9 +38,17 @@ const schema = {
       _type: Type.ConceptUuid,
       _default: "5089AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
     },
-    respiratoryRate: {
+    respiratoryRateUuid: {
```
brandones, line 41: This is a good fix though, thank you :+1:

__mocks__/chart-widgets-config.mock.ts
jonathandick: I don't know the the are coded via metadata. I think they are part of the concept itself. Such that weight in lb's would use a a different than weight in kgs. I just looked in the openmrs concept dictionary and there doesn't seem to be any way to infer units from these concepts other than to just "know" what the right one is.  @donaldkibet you probably have a dictionary object that maps concepts to their corresponding units. If someone were to use the weight concept (for example), you would be displaying the wrong units.

src/widgets/vitals/vitals-biometrics-form/vitals-biometrics-form.component.tsx
```diff
@@ -0,0 +1,375 @@
...
+  concepts: {
+    systolicBloodPressureUuid: { max: 120, min: 110 },
+    diastolicBloodPressureUuid: { max: 80, min: 75 },
+    heartRateUuid: { max: 100, min: 60 },
+    temperatureUuid: { max: 37.5, min: 36.5 },
+    spo2Uuid: { max: 100, min: 95 },
+    respiratoryRateUuid: { max: 20, min: 12 },
+    bmi: { max: 24.9, min: 18.5 },
+    midUpperArmCircumference: { max: 25.5, min: 23 }
+  }
```
brandones: These, too, should come from the concepts

__mocks__/chart-widgets-config.mock.ts
brandones: It's definitely part of the model:

screenshot--2020-12-08--102332

jonathandick: I stand corrected

src/widgets/vitals/vitals-biometrics-form/vitals-biometrics-form.component.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+    userEvent.type(screen.getByTitle(/diastolic/i), "80");
+    userEvent.type(screen.getByTitle(/Notes/i), "Patient vitals okay");
+    userEvent.type(screen.getByTitle(/Weight/i), "70");
+    userEvent.type(screen.getByTitle(/Height/i), "165");
+    userEvent.type(screen.getByTitle(/MUAC/i), "28");
+    mockSavePatientVitals.mockImplementation(() =>
+      Promise.resolve({ status: 201 })
+    );
+    userEvent.click(saveButton);
+    expect(mockSavePatientVitals).toHaveBeenCalled();
```
brandones: I think asserting on the arguments with which this function was called, i.e.,
`expect(mockSavePatientVitals).toHaveBeenCalledWith()`, would make this a very valuable test. Right now it is not very valuable.

Note that you [don't even have to assert on all the parameters](https://codewithhugo.com/jest-specific-argument-parameter-assert/). The really important argument is the one that contains the data. Testing that will very concisely test a whole lot of logic distributed throughout the component.

src/widgets/vitals/vitals-biometrics-form/vitals-biometrics-form.utils.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const isInValidRange = (
+  conceptsMetaData: Array<ConceptMetaData>,
+  conceptUuid: string,
+  value: string | number
+) => {
+  if (value === undefined || value === "") return false;
+  const normalRange = conceptsMetaData.find(c => c.uuid === conceptUuid);
+  if (!normalRange?.lowNormal || !normalRange?.hiNormal) return false;
+  return !(value >= normalRange?.hiNormal && value <= normalRange?.lowNormal);
```
brandones, line 17:
```suggestion
  const concept = conceptsMetaData.find(c => c.uuid === conceptUuid);
  if (!concept?.lowNormal || !concept?.hiNormal) return false;
  return !(value >= concept?.hiNormal && value <= concept?.lowNormal);
```

brandones: Please write a test for this function. I do not think it works correctly. Might as well write one for `calculateBMI` too.

brandones: Actually, instead of writing a test for this function, I suppose it would be better to write tests for VitalsAndBiometricForms which test the validation behavior (and the BMI calculation). Per the advice "[avoid testing implementation details](https://kentcdodds.com/blog/testing-implementation-details)." That way the tests will be robust to refactoring and other "implementation detail" changes.

brandones (changes requested):
Great progress. Just a couple more changes needed regarding tests for the form, and the behavior of the validation function.

src/widgets/vitals/vitals-biometrics-form/vitals-biometrics-form.utils.ts
```diff
@@ -0,0 +1,24 @@
...
+  if (value === undefined || value === "") return false;
+  const normalRange = conceptsMetaData.find(c => c.uuid === conceptUuid);
+  return isNumber(normalRange.hiNormal) && isNumber(normalRange.lowNormal)
+    ? !inRange(
+        Number(value),
+        Number(normalRange.lowNormal),
+        Number(normalRange.hiNormal)
+      )
+    : false;
+};
```
brandones: I think this function is upside-down. It's called `isInValidRange`, and it returns `true` when the value is bad, and `false` when the value is good (or there's no applicable range).

Also is about "normal," not "valid".

src/widgets/vitals/vitals-biometrics-form/vitals-biometrics-input.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+const VitalsBiometricInput: React.FC<VitalsBiometricInputProps> = ({
+  title,
+  onInputChange,
+  textFields,
+  unitSymbol,
+  textFieldStyles,
+  textFieldWidth,
+  placeholder,
+  disabled,
+  isValidRange: isValid
```
brandones:
```suggestion
  inputIsValid
```

src/widgets/vitals/vitals-biometrics-form/vitals-biometrics-input.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+}) => {
+  return (
+    <div className={styles.inputContainer} style={{ width: textFieldWidth }}>
+      <p className={styles.vitalsBiometricInputLabel01}>{title}</p>
+      <div
+        className={`${styles.textInputContainer} ${disabled &&
+          styles.disableInput} ${isValid && styles.danger}`}
```
brandones:
```suggestion
          styles.disableInput} ${!inputIsNormal && styles.danger}`}
```

src/widgets/vitals/vitals-biometrics-form/vitals-biometrics-input.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+        style={{ ...textFieldStyles }}
+      >
+        <div className={styles.centerDiv}>
+          {textFields.map(val => {
+            return val.type === "text" ? (
+              <Fragment key={val.name}>
+                <TextInput
+                  style={{ ...textFieldStyles }}
+                  className={`${styles.textInput} ${disabled &&
+                    styles.disableInput} ${val.className} ${isValid &&
```
brandones:
```suggestion
                    styles.disableInput} ${val.className} ${!inputIsNormal &&
```

src/widgets/vitals/vitals-biometrics-form/vitals-biometrics-form.component.tsx
```diff
@@ -0,0 +1,383 @@
...
+                  value: patientVitalAndBiometrics?.systolicBloodPressure || ""
+                },
+                {
+                  name: t("diastolic", "diastolic"),
+                  type: "text",
+                  value: patientVitalAndBiometrics?.diastolicBloodPressure || ""
+                }
+              ]}
+              unitSymbol={bloodPressureUnit}
+              isValidRange={
```
brandones:
```suggestion
              inputIsNormal={
```

etc for all the other usages in this file

brandones (approved):
After the suggestions I've made are applied—LGTM!

src/widgets/vitals/vitals-biometrics-form/vitals-biometrics-input.component.tsx
brandones: This is the change corresponding to the backwards function implementation. Right now this line says "if the input is valid, mark it with danger." What we want is "if the input is not valid, mark it with danger."

src/widgets/vitals/vitals-biometrics-form/vitals-biometrics-form.component.tsx
```diff
@@ -0,0 +1,383 @@
...
+    pulseUnit,
+    oxygenSaturationUnit,
+    midUpperArmCircumferenceUnit,
+    respiratoryRateUnit
+  ] = conceptsUnits;
+
+  const isBMIInValidRange = value => {
+    if (value === undefined || value === "") return false;
+    return !(value >= 18.5 && value <= 24.9);
+  };
```
brandones: This implementation is also backwards

brandones (changes requested):
Oh whoops, I'm just realizing that all of this code uses the word "valid" when it is actually about "normal." Those are two completely different things. Please correct all the names.

src/widgets/vitals/vitals-biometrics-form/vitals-biometrics-form.utils.ts
brandones:
```suggestion
export const isInNormalRange = (
  conceptsMetaData: Array<ConceptMetaData>,
  conceptUuid: string,
  value: string | number
) => {
  if (value === undefined || value === "") return true;
  const concept = conceptsMetaData.find(c => c.uuid === conceptUuid);
  return isNumber(concept.hiNormal) && isNumber(concept.lowNormal)
    ? inRange(
        Number(value),
        Number(concept.lowNormal),
        Number(concept.hiNormal)
      )
    : true;
};
```

src/widgets/vitals/vitals-biometrics-form/vitals-biometrics-input.component.tsx
brandones:
```suggestion
  inputIsNormal
```

src/widgets/vitals/vitals-biometrics-form/vitals-biometrics-form.component.tsx
brandones:
```suggestion
  const isBMIInNormalRange = value => {
    if (value === undefined || value === "") return true;
    return value >= 18.5 && value <= 24.9;
  };
```

brandones (approved):
Looks good! Thanks!

PR #166, opened 2020-12-09
MF-385: Carbonise overview widgets

denniskigen:
This PR rewrites the patient chart summary overview widgets to match the new designs.

Screenshot from 2020-12-09 22-47-42

Screenshot from 2020-12-09 22-48-19

**TODO**: Pagination. Will follow in a future PR.


src/widgets/notes/encounter.resource.tsx
```diff
@@ -54,3 +56,30 @@ export function fetchEncounterByUuid(encounterUuid): Observable<any> {
...
+function formatNotes(notes: Array<RESTPatientNote>): Array<PatientNote> {
+  let formattedNotes: Array<PatientNote> = [];
+  notes.forEach((note: RESTPatientNote) => {
+    formattedNotes.push(mapNoteProperties(note));
+  });
+  return formattedNotes;
+}
+
+function mapNoteProperties(note: RESTPatientNote): PatientNote {
+  const formattedNote: PatientNote = {
```
FlorianRappl: Just return:

```ts
function mapNoteProperties(note: RESTPatientNote): PatientNote {
  return {
    ...
  };
}

File: src/widgets/notes/encounter.resource.tsx [hunk 2]
```diff
@@ -54,3 +56,30 @@ export function fetchEncounterByUuid(encounterUuid): Observable<any> {
     map(({ data }) => data)
   );
 }
+
+function formatNotes(notes: Array<RESTPatientNote>): Array<PatientNote> {
+  let formattedNotes: Array<PatientNote> = [];
```
FlorianRappl: return notes.map(mapNoteProperties);

src/widgets/notes/encounter.resource.tsx
```diff
@@ -54,3 +56,30 @@ export function fetchEncounterByUuid(encounterUuid): Observable<any> {
...
+  notes.forEach((note: RESTPatientNote) => {
+    formattedNotes.push(mapNoteProperties(note));
+  });
+  return formattedNotes;
+}
+
+function mapNoteProperties(note: RESTPatientNote): PatientNote {
+  const formattedNote: PatientNote = {
+    id: note.uuid,
+    encounterDate: note.encounterDatetime.slice(0, 19), // TODO: FIX THIS BAD HACK
```
FlorianRappl: Why 19? (4+1+2+1+2=10) + 1 + (2+1+2+1+2=8), like `YYYY-MM-DDTHH:MM:SS`? Why then stop? What's afterwards? If its ISO its the offset (e.g., `Z` for UTC, otherwise some offset relative to UTC). That would be relevant information for parsing / showing it.

src/widgets/notes/notes-overview.component.tsx
```diff
@@ -1,106 +1,164 @@
...
+      key: "encounterLocation",
+      header: t("location", "Location")
+    },
+    {
+      key: "encounterAuthor",
+      header: t("author", "Author")
+    }
+  ];
+
+  const getRowItems = rows =>
```
FlorianRappl: Any reason for capturing this function or could we define it outside, too? What (type) is `rows`?

src/widgets/notes/notes-overview.component.tsx
```diff
@@ -1,106 +1,164 @@
...
+    rows.map(row => ({
+      ...row,
+      encounterDate: formatNotesDate(row.encounterDate),
+      name: row.encounterName,
+      location: row.encounterLocation,
+      author: row.encounterAuthor ? row.encounterAuthor : "\u2014"
+    }));
+
+  const RenderNotes = () => {
```
FlorianRappl: Is this `React.FC`?

FlorianRappl (commented):
Great work!

src/widgets/notes/encounter.resource.tsx
denniskigen: Yes, it's an ISO date string (with UTC offset). Fixed.

src/widgets/notes/notes-overview.component.tsx
denniskigen: In this case, the same as the `notes`, so type `PatientNote`. I defined it locally so I can make manipulations to the data coming in from the `note` object before it gets rendered. 

```
  const getRowItems = (rows: Array<PatientNote>) =>
    rows.map(row => ({
      ...row,
      encounterDate: formatNotesDate(row.encounterDate),
      author: row.encounterAuthor ? row.encounterAuthor : "\u2014"
    }));
```

src/widgets/notes/notes-overview.component.tsx
denniskigen: Yes!

FlorianRappl: const RenderNotes: React.FC = ... ;)

src/widgets/notes/notes-overview.component.tsx
FlorianRappl: Yeah what I'm after is that we could have defined that in a function outside (thus making the render more lightweight).

```ts
function getRowItems(rows: Array<PatientNote>) {
  return rows.map(row => ({ ...row, ... }));
}
```

Reason is that `formatNotesDate` was available outside, so I did not see any captured value.

src/widgets/notes/notes-overview.component.tsx
denniskigen: Fixed

src/widgets/notes/notes-overview.component.tsx
denniskigen: Ah! Very clear. Thanks

jonathandick:
@denniskigen , looks like you've addressed all the issues. Could you fix conflicts and then we can merge?

denniskigen:
Up to date.

PR #167, opened 2020-12-11
Fix subtle bug in allergies resource

denniskigen:
This PR

- Fixes a bug in the `performPatientAllergySearch` function when an empty result is returned from the API request for allergies.
- Also removes a superfluous `console.log` from a test file.


brandones (approved):
Thanks Dennis!

PR #171, opened 2020-12-15
MF-398: Onset date should not be required

denniskigen:
[Ticket](https://issues.openmrs.org/browse/MF-398)

`Onset date` should be an optional field when recording a condition.


ibacher (approved):

ibacher:
LGTM

PR #172, opened 2020-12-15
MF-383 : Implemented carbon styled vitals chart view

donaldkibet:
### What does this PR do?

It implements vitals widget chart view

### Screenshot
vitals-chart

### DIFF
--- a/src/widgets/vitals/vitals-chart.component.tsx
+++ b/src/widgets/vitals/vitals-chart.component.tsx
@@ -1,0 +1,166 @@
+import React from "react";
+import styles from "./vitals-chart.component.scss";
+import "@carbon/charts/styles.css";
+import { RadioButtonGroup, RadioButton } from "carbon-components-react";
+import { PatientVitals } from "./vitals-biometrics.resource";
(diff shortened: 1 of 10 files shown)


jonathandick:
@nmalyschkin are you able to review?

src/widgets/vitals/vitals-chart.component.tsx
```diff
@@ -0,0 +1,166 @@
...
+  React.useEffect(() => {
+    const chartData = patientVitals
+      .map(vitals => {
+        return {
+          group: selectedVitalSign.groupName,
+          key: dayjs(vitals.date).format("DD-MMM"),
+          value: vitals[selectedVitalSign.value]
+        };
+      })
+      .sort((a: any, b: any) => b.date - a.date);
```
nmalyschkin: Seems to me that this `sort` would have no effect, since there is no `date` attribute. Was the sort operation supposed to be done before the mapping?

nmalyschkin (changes requested):

jonathandick:
I"m going to merge this as the issue appears to have been addressed by latest commit.

PR #174, opened 2020-12-23
MF-390: Rewrite Visit Notes form based on new designs

denniskigen:
[Ticket](https://issues.openmrs.org/browse/MF-390)

This PR rewrites the `VisitNotesForm` component to match the new [designs](https://app.zeplin.io/project/5f7223cfda10f94d67cec6d0/screen/5fa44075a372569d4f07f968). 

Screenshot 2020-12-23 at 18 13 36

Still TODO:

- Handle uploading of images.
- Handle form validation.
- Seek clarity on the functionality of the search field - from the designs it appears it was intended to work like a  filter where the user input would be checked against a (possibly) pre-fetched list of diagnoses. AFAICT, the api exposes just a `search` method for retrieving diagnoses - suggesting that pre-fetching ALL diagnoses ahead of time is not supported. I elected to implement it as a simple search pending clarification.


jonathandick:
The search should hit the spot and not be a drop down list

brandones:
@denniskigen what's the status of this? Is it ready for review?

denniskigen:
Yep, ready for review @brandones

src/widgets/notes/visit-notes-form.component.tsx
```diff
@@ -0,0 +1,380 @@
...
+    );
+  };
+
+  interface DiagnosisSearchResultsProps {
+    results: Array<Diagnosis>;
+  }
+
+  const DiagnosisSearchResults: React.FC<DiagnosisSearchResultsProps> = ({
+    results
+  }) => <>{results ? <RenderSearchResults /> : <SearchSkeleton />}</>;
```
brandones, line 196: Let's not create components in the body of other components. This can just me moved above or below (but out of) VisitNotesForm.

brandones: Let's not create components in the body of other components. This can just me moved above or below (but out of) VisitNotesForm.

src/widgets/notes/visit-notes-form.component.tsx
```diff
@@ -0,0 +1,380 @@
...
+
+  const handleSubmit = $event => {
+    $event.preventDefault();
```
brandones:
```suggestion
  const handleSubmit = event => {
    event.preventDefault();
```

src/widgets/notes/visit-notes-form.component.tsx
```diff
@@ -0,0 +1,380 @@
...
+                placeHolderText={t(
+                  "diagnosisInputPlaceholder",
+                  "Choose primary diagnosis first, then secondary diagnoses"
+                )}
+                onChange={e =>
+                  handleSearchTermChange(e.currentTarget.value ?? "")
+                }
+                ref={searchInputRef}
+              />
+              {!!searchTerm && (
```
brandones:
```suggestion
              {searchTerm && (
```

src/widgets/types.ts
```diff
@@ -304,3 +304,65 @@ type DisplayMetadata = {
...
+  retired: boolean;
+  links: Links;
+  resourceVersion: string;
+}
+
+export interface Location {
+  uuid: string;
+  display: string;
+  name: string;
+  description: string | null;
```
brandones: These come in as null rather than undefined? Otherwise just use `?:`.

src/widgets/visit/visit-button.component.tsx
```diff
@@ -132,10 +132,8 @@ export const StartVisitConfirmation: React.FC = () => {
   return (
     <div className={styles.visitPromptContainer}>
       <h2>
-        {t(
-          "startVisitPrompt",
-          "No active visit is selected. Do you want to start a visit?"
-        )}
+        No active visit is selected. <br />
+        Do you want to start a visit?
```
brandones: What happened to the translation?

brandones (changes requested):
Really strong work. Just a couple minor thing to fix: no components within components, no unnecessary boolean cast, and a lost translation.

src/widgets/visit/visit-button.component.tsx
denniskigen: This change shouldn't be here. Reverted.

src/widgets/notes/visit-notes-form.component.tsx
```diff
@@ -0,0 +1,371 @@
...
+
+  const handleRemoveDiagnosis = (diagnosisToRemove: Diagnosis) => {
+    setSelectedDiagnoses(
+      selectedDiagnoses.filter(
+        diagnosis => diagnosis.concept.id !== diagnosisToRemove.concept.id
+      )
+    );
+  };
+
+  const RenderSearchResults: React.FC = () => {
```
brandones: Another component defined within a component. Either leave this stuff inline in the `render` block or factor it all the way out of the component (with `searchResults`, `handleAddDiagnosis`, and whatever else as props).

brandones (changes requested):

src/widgets/notes/visit-notes-form.component.tsx
denniskigen: Sorry @brandones, refactored this while tired. Fixed.

brandones: No worries, thanks!

brandones (approved):
LGTM! Thanks @denniskigen !

PR #176, opened 2020-12-29
Fix immunization widget config schema

brandones:
I should add a validator to esm-config to catch this kind of error. `_type` should be present at the same level as `_default`.

CC @mddubey


jonathandick (approved):
lgtm

PR #178, opened 2021-01-11
[MF-401] Add patient-chart menu items as extensions

samuelmale:
Ticket: https://issues.openmrs.org/browse/MF-401

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -97,6 +97,92 @@ function setupOpenMRS() {
             ),
           { featureName: "vitals-biometric-form", moduleName }
         )
+      },
+      {
+        id: "summary-menu-item",
+        slot: "patient-chart-nav-menu",
+        load: getAsyncLifecycle(() => import("./menu-items/summary-link"), {
+          featureName: "summary-menu-item",
+          moduleName
+        })
+      },
+      {
+        id: "attachments-menu-item",
+        slot: "patient-chart-nav-menu",
+        load: getAsyncLifecycle(() => import("./menu-items/attachments-link"), {
+          featureName: "attachments-menu-item",
+          moduleName
+        })
+      },
+      {
+        id: "results-menu-item",
+        slot: "patient-chart-nav-menu",
+        load: getAsyncLifecycle(() => import("./menu-items/results-link"), {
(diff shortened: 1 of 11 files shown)


FlorianRappl (approved):
LGTM - thanks :beers:!

src/index.ts
```diff
@@ -97,6 +97,92 @@ function setupOpenMRS() {
             ),
           { featureName: "vitals-biometric-form", moduleName }
         )
+      },
```
jonathandick, line 100: @samuelmale these items should be read from a configuration not just hard coded into the app.

brandones: @jonathandick They are configurable by virtue of being extensions. This all seems correct to me.

src/menu-items/attachments-link.tsx
```diff
@@ -0,0 +1,16 @@
+import { ConfigurableLink } from "@openmrs/esm-react-utils";
+import React from "react";
+import { useTranslation } from "react-i18next";
+
+export default function({ basePath }) {
+  const { t } = useTranslation();
+
+  return (
+    <ConfigurableLink
+      to={`${basePath}/attachments`}
```
jonathandick, line 10: @samuelmale it looks like the basePath itself must include the last "/" as the path contains "//". 
image 

This is true for all the links you've created.

brandones: ~Uh-oh! ConfigurableLink is supposed to already deal with that intelligently. @samuelmale is using this correctly. I'll look into it.~

Wait, I misunderstood what's going on here. You're correct. This has nothing to do with ConfigurableLink behavior.

samuelmale: @jonathandick this is happening because the `basePath` has an extra forward slash which I fixed [here](https://github.com/openmrs/openmrs-esm-patient-chart/pull/204/files#diff-fdcb51c6cb31fcfecd78b958740498466c84e2f38ff9e47a1056e74c44f4a582R12-R14) but the associated PR https://github.com/openmrs/openmrs-esm-patient-chart/pull/204 isn't merged in yet. Seems @FlorianRappl just forgot about it.

PR #181, opened 2021-02-01
[MF-405] Add widget as extension to capture patient photo

samuelmale:
Ticket: https://issues.openmrs.org/browse/MF-405

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -196,10 +196,10 @@ function setupOpenMRS() {
         )
       },
       {
-        id: "capture-patient-photo",
+        id: "capture-photo",
         slot: "capture-patient-photo",
         load: getAsyncLifecycle(
-          () => import("./widgets/attachments/capture-patient-photo.component"),
+          () => import("./widgets/attachments/capture-photo.component"),
           {
             featureName: "allergies-menu-item",
             moduleName
--- a/src/index.ts
+++ b/src/index.ts
@@ -194,6 +194,17 @@ function setupOpenMRS() {
             moduleName
           }
         )
+      },
+      {
+        id: "capture-patient-photo",
+        slot: "capture-patient-photo",
+        load: getAsyncLifecycle(
+          () => import("./widgets/attachments/capture-patient-photo.component"),
+          {
+            featureName: "allergies-menu-item",
+            moduleName
+          }
+        )
       }
     ]
   };
--- a/src/index.ts
+++ b/src/index.ts
@@ -201,7 +201,7 @@ function setupOpenMRS() {
         load: getAsyncLifecycle(
           () => import("./widgets/attachments/capture-photo.component"),
           {
-            featureName: "allergies-menu-item",
+            featureName: "capture-photo-widget",
             moduleName
           }
         )
--- a/src/widgets/attachments/camera-frame.component.tsx
+++ b/src/widgets/attachments/camera-frame.component.tsx
@@ -41,7 +41,6 @@ export default function CameraFrame(props: CameraFrameProps) {
 
 type CameraFrameProps = {
   children: React.ReactNode;
-  onCloseCamera: Function;
-  onSelectFile?: Function;
-  setSelectedFile?: Function;
+  onCloseCamera(): void;
+  setSelectedFile?(file: File): void;
 };
--- a/src/widgets/attachments/camera-frame.component.tsx
+++ b/src/widgets/attachments/camera-frame.component.tsx
@@ -1,6 +1,5 @@
 import React from "react";
 import styles from "./camera-frame.css";
-import attachmentsOverviewStyles from "./attachments-overview.css";
 
 export default function CameraFrame(props: CameraFrameProps) {
   function handleClick() {
@@ -26,7 +25,13 @@ export default function CameraFrame(props: CameraFrameProps) {
             <label htmlFor="uploadPhoto" className={styles.choosePhoto}>
               Select photo
             </label>
-            <input type="file" id="uploadPhoto" style={{ display: "none" }} />
+            <input
+              type="file"
+              id="uploadPhoto"
+              accept="image/*"
+              style={{ display: "none" }}
+              onChange={e => props.setSelectedFile(e.target.files[0])}
+            />
           </form>
         </div>
       </div>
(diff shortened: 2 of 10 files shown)


src/index.ts
```diff
@@ -194,6 +194,17 @@ function setupOpenMRS() {
             moduleName
           }
         )
+      },
+      {
+        id: "capture-patient-photo",
```
FlorianRappl: This is way too specific.

So it sounds like this has a single use-case. Shouldn't it be a generic upload / attachment / make photo widget?

src/widgets/attachments/camera-frame.component.tsx
```diff
@@ -26,4 +42,6 @@ export default function CameraFrame(props: CameraFrameProps) {
 type CameraFrameProps = {
   children: React.ReactNode;
   onCloseCamera: Function;
+  onSelectFile?: Function;
```
FlorianRappl: What is `Function` (also above and below)? Can we be more specific?

src/widgets/attachments/camera-upload.component.tsx
```diff
@@ -79,5 +116,11 @@ export default function CameraUpload(props: CameraUploadProps) {
 }
 
 type CameraUploadProps = {
-  onNewAttachment: Function;
+  openCameraOnRender?: Boolean;
```
FlorianRappl: It's `boolean` and not `Boolean`. Same for `string` and `number`.

src/widgets/attachments/camera-upload.component.tsx
```diff
@@ -79,5 +116,11 @@ export default function CameraUpload(props: CameraUploadProps) {
...
+  shouldNotRenderButton?: Boolean;
+  closeCamera?: Function;
+  onTakePhoto?: Function;
+  delegateSaveImage?: Function;
+  selectedFile?: File;
+  onNewAttachment?: Function;
```
FlorianRappl: Can we be more specific on these `Function`s?

FlorianRappl (commented):
I think we are on a good way here! Thanks :beers:

src/index.ts
samuelmale: This is an extension that renders a component that is specific to capturing the patient's photo([CapturePatientPhoto](https://github.com/openmrs/openmrs-esm-patient-chart-widgets/blob/f883d440a93e659f8d4d17f2a71d37bddf7dbac4/src/widgets/attachments/capture-patient-photo.component.tsx#L33)). This component does more heavy lifting for the feature; like providing the associated concept complex, obs date in desired backend format. The registration component only implements the POST operation. However I can work on a followup commit to make this component more generic. The [concept](https://github.com/openmrs/openmrs-esm-patient-chart-widgets/blob/f883d440a93e659f8d4d17f2a71d37bddf7dbac4/src/widgets/attachments/capture-patient-photo.component.tsx#L33) can be configured in the registration esm?

FlorianRappl: Well, this is patient-chart-widgets here; I mean how often (besides registration) will we need to take the patient's photo? Once? Twice? More often we'll take *some* photo.

What we could do is to leave this special variant here, but also provide (and use in the special variant) a more generic version.

samuelmale: >Well, this is patient-chart-widgets here; I mean how often (besides registration) will we need to take the patient's photo? Once? Twice? More often we'll take some photo.


The component that captures 'a photo' is existing and purely generic:- [CameraUpload](https://github.com/samuelmale/openmrs-esm-patient-chart-widgets/blob/MF-405/src/widgets/attachments/camera-upload.component.tsx#L13), you can see that `CapturePatientPhoto` component references it [here](https://github.com/samuelmale/openmrs-esm-patient-chart-widgets/blob/MF-405/src/widgets/attachments/capture-patient-photo.component.tsx#L54-L59) with a configuration suitable for the patient photo feature.

>What we could do is to leave this special variant here, but also provide (and use in the special variant) a more generic version.

Okay, I agree that we really need to improve these components in regards to re-usability.

FlorianRappl: I think we have a disconnect here. `CameraUpload` is a component yes, but I refer to it / want / need it to be (usable as) an extension. Otherwise some microfrontend cannot leverage this.

samuelmale: I get what you mean, so let me work on making the `CapturePatientPhoto` more generic :)

samuelmale: Updated PR, also opened a followup PR on Reg. --> https://github.com/openmrs/openmrs-esm-patient-registration/pull/65

FYI: In my recent commit, I also improved on the styling 👇

samuelmale: Oooppss, let me scrap-off the config ie. `config.concepts.patientPhotoUuid`

samuelmale: >Oooppss, let me scrap-off the config ie. config.concepts.patientPhotoUuid


Done ☺️

src/widgets/attachments/image-preview.component.tsx
```diff
@@ -38,24 +39,33 @@ export default function ImagePreview(props: ImagePreviewProps) {
...
+          type="text"
+          placeholder={t(
+            "attachmentCaptionInstruction",
+            "Enter a caption for the image"
+          )}
+          onChange={updateCaption}
+        />
+      )}
+      <ButtonSet style={{ width: "50%" }}>
+        <UserHasAccess privilege="Create Attachment">
```
FlorianRappl: If the only thing we can do is to cancel - why have something at all here?

Shouldn't the whole button set / form be hidden if no privileges are there?

FlorianRappl (commented):
LGTM! :beers:

src/widgets/attachments/image-preview.component.tsx
samuelmale: Resolved.

src/index.ts
```diff
@@ -194,6 +194,17 @@ function setupOpenMRS() {
...
           }
         )
+      },
+      {
+        id: "capture-photo",
+        slot: "capture-patient-photo",
+        load: getAsyncLifecycle(
+          () => import("./widgets/attachments/capture-photo.component"),
+          {
+            featureName: "allergies-menu-item",
```
FlorianRappl: The feature name should be "capture-photo-widget" or so, right?

FlorianRappl (commented):
I think we are close to the finish line here. Thanks! 🏖️

src/index.ts
samuelmale: Nice catch, this was an oversight. Updated!

FlorianRappl (approved):
Thanks!

PR #182, opened 2021-02-02
Align tile states text content explicitly

denniskigen:
Screenshot 2021-02-02 at 14 08 03

Screenshot 2021-02-02 at 14 07 56

The text inside the empty and error state tiles is mistakenly receiving horizontal alignment from a class declared in the `patient chart` app. This commit adds explicit horizontal alignment to the tiles instead.


FlorianRappl (approved):
Great! :beers:

PR #193, opened 2021-02-09
MF-450: Active Visit tag should be visible when there is an active visit

ibacher:
Ticket: https://issues.openmrs.org/browse/MF-450

Screenshot showing Active Visit next to a patient with an active visit

### DIFF
--- a/src/widgets/banner/patient-banner.component.tsx
+++ b/src/widgets/banner/patient-banner.component.tsx
@@ -1,24 +1,45 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 
 import dayjs from "dayjs";
-import { Button } from "carbon-components-react";
+import { Button, Tag } from "carbon-components-react";
 import { CaretDown16, CaretUp16 } from "@carbon/icons-react";
 import capitalize from "lodash-es/capitalize";
 
 import { useCurrentPatient } from "@openmrs/esm-react-utils";
+import { useTranslation } from "react-i18next";
 
 import ContactDetails from "../contact-details/contact-details.component";
 
 import placeholder from "../../assets/placeholder.png";
 import { age } from "../contact-details/age-helpers";
 import styles from "./patient-banner.scss";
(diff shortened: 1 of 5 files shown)


FlorianRappl (approved):
LGTM - thanks :beers:!

src/widgets/banner/patient-banner.component.tsx
```diff
@@ -1,25 +1,46 @@
...
+  useEffect(() => {
+    if (patient) {
+      const abortController = new AbortController();
+      const sub = getVisitsForPatient(
+        patient.id,
+        abortController,
+        "custom:(stopDatetime)"
+      ).subscribe(({ ok, data }) => {
+        if (ok) {
+          setHasActiveVisit(data.results?.some(v => v.stopDatetime == null));
```
jonathandick, line 36: This warrants some discussion - it looks like if a patient has any visit that has not been ended that you are showing it as an active visit. Should we be more selective? Perhaps only if the most recent visit is active? Alternatively, it will be possible to edit/add to a retrospective visit - how will you accommodate this? In general. I think we may want to have a single point to set/test if the visit is active. @brandones what do you think?

ibacher: This is actually slightly stricter than the notion of "active visits"  in the backend (see [here](https://github.com/openmrs/openmrs-core/blob/c5001a75d55d519fb6433f512d438f3c27ca163d/api/src/main/java/org/openmrs/api/db/hibernate/HibernateVisitDAO.java#L200) and the other implementation [here](https://github.com/openmrs/openmrs-module-emrapi/blob/a8da7c96d3649788d8db522ba7274b28cc2cf42c/api/src/main/java/org/openmrs/module/emrapi/visit/VisitDomainWrapper.java#L159)), but I agree that the points here are worth discussing.

brandones: Yeah I haven't thought much about it, but I think if the backend has a notion of what an "active visit" is, then we should respect that.

PR #194, opened 2021-02-09
Fixes some unit tests and adds some appropriate typing

ibacher:
This PR does two things:

1. It fixes some tests that are broken for me due to timezone differences. I'm in UTC-0500 currently; a few tests rely on DayJS converting strings given in UTC to particular days; this works well if you are in UTC or UTC+<something>, but fails in UTC-<something>. For example, one test relied on the string "2020-10-01T00:00:00+00:00" converting to "Oct-2020" (using the "MMM-YYYY" format), but, of course, for my locale, it renders as "Sep-2020" because 2020-10-01T00:00:00+00:00 is 2020-09-30T19:00:00-05:00.

2. I added some types to the program-form which were necessary to do the above cleanly.


brandones (approved):
Yesssssss I've been hoping someone would fix the time zone issue with the tests for a long time. Thanks so much!

PR #195, opened 2021-02-10
MF-450: Active Visit tag should be visible when there is an active visit

ibacher:
Simpler implementation of the Active Visit tag that only displays when a visit is selected.

### DIFF
--- a/src/widgets/banner/patient-banner.component.tsx
+++ b/src/widgets/banner/patient-banner.component.tsx
@@ -13,33 +13,24 @@ import ContactDetails from "../contact-details/contact-details.component";
 import placeholder from "../../assets/placeholder.png";
 import { age } from "../contact-details/age-helpers";
 import styles from "./patient-banner.scss";
-import { getVisitsForPatient } from "../visit/visit.resource";
+import { getStartedVisit, visitItem, visitMode } from "../visit/visit-utils";
 
 export default function PatientBanner() {
   const [showContactDetails, setShowContactDetails] = useState(false);
-  const [hasActiveVisit, setHasActiveVisit] = useState(false);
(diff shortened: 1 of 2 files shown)


src/widgets/banner/patient-banner.component.tsx
```diff
@@ -13,33 +13,24 @@ import ContactDetails from "../contact-details/contact-details.component";
 import placeholder from "../../assets/placeholder.png";
 import { age } from "../contact-details/age-helpers";
 import styles from "./patient-banner.scss";
-import { getVisitsForPatient } from "../visit/visit.resource";
+import { getStartedVisit, visitItem, visitMode } from "../visit/visit-utils";
```
jonathandick, line 16: If we store the "active visit" in patient chart, and the user logs out, logs back in and returns to the same patient chart, do they now need to set again as active? Should we be saving this in localStorage as well? @FlorianRappl @brandones

brandones: "Yes" and "yes probably but other suggestions are welcome."

ibacher: > returns to the same patient chart, do they now need to set again as active?

Or if they just reload the page, which is a decent argument for using localStorage or something similar. We also need some kind of contextual boundaries. Activating a visit on one patient and then switching to another patient seems to leave the visit as active.

brandones: Yeah my suggested implementation at this point is that `activeVisits` should be a piece of state that ser/des to localStorage, and which just stores the active visit for each patient along with a datestamp. The datestamp is used to expire entries more than e.g. 24h old.

PR #198, opened 2021-02-16
Implemented Forms with, with ability to launch form-entry  form

donaldkibet:
### What does this PR do? 

1. Add ability to launch form-entry forms
2. Implements the form-widget as shown in the designs [ticket](https://app.zeplin.io/project/5f7223cfda10f94d67cec6d0/screen/60182928c320622924b940ff)

### To do
1. Write tests

### DIFF
--- a/src/widgets/forms/forms-utils.ts
+++ b/src/widgets/forms/forms-utils.ts
@@ -1,0 +1,32 @@
+import { Encounter, Form } from "../types";
+
+export function filterAvailableAndCompletedForms(
+  forms: Array<Form>,
+  encounters: Array<Encounter>
+): {
+  available: Array<Form>;
+  completed: Array<Encounter>;
+} {
+  const availability = {
+    available: [],
+    completed: []
+  };
+
+  forms.forEach(form => {
+    let completedEncounters = encounters.filter(encounter => {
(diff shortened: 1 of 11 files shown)


src/widgets/forms/forms-utils.ts
```diff
@@ -0,0 +1,32 @@
...
+  available: Array<Form>;
+  completed: Array<Encounter>;
+} {
+  const availability = {
+    available: [],
+    completed: []
+  };
+
+  forms.forEach(form => {
+    let completedEncounters = encounters.filter(encounter => {
```
FlorianRappl: const, right?

src/widgets/forms/forms-utils.ts
```diff
@@ -0,0 +1,32 @@
...
+      return areFormsEqual(encounter.form, form);
+    });
+    if (completedEncounters.length > 0) {
+      availability.completed = availability.completed.concat(
```
FlorianRappl: Why not use `push` here, too? You can then just `...` it.

FlorianRappl (commented):
LGTM - thanks!

FlorianRappl (approved):
Perfect - thanks :beers:!

PR #203, opened 2021-02-22
MF-442: Add ability to expand alert header by on the alert

donaldkibet:
### ticket

[MF-442](https://issues.openmrs.org/projects/MF/issues/MF-442?filter=allissues)

### What does this PR do?

1. Add ability to expand and collapse alert header by clicking anywhere on the alert
2. remove config as a dependency as it cause un-necessary re-render


FlorianRappl (approved):
LGTM - thanks a lot :beers:!

PR #205, opened 2021-02-23
Add ability to require a visit before opening forms

donaldkibet:
### What does this PR do?

1. Require to have an active visit to open a form and update the active visits correctly

### Screenshoot

screencast 2021-02-23 14-56-07


jonathandick:
@donaldkibet could you fix conflicts? I think we can then merge this. This is a temp fix and will likely need to be replaced with the official designs once available.

PR #207, opened 2021-03-05
[test-results] add loadPatientData and Overview component

nmalyschkin:
This is the first chunk of existing code

### DIFF
--- a/src/widgets/test-results/loadPatientTestData/helpers.ts
+++ b/src/widgets/test-results/loadPatientTestData/helpers.ts
@@ -1,0 +1,250 @@
+import { getGlobalStore } from "@openmrs/esm-state";
+import {
+  PatientData,
+  ObsRecord,
+  ConceptUuid,
+  ConceptRecord,
+  ObsMetaInfo
+} from "./types";
+
+const retrieveFromIterator = <T>(
+  iteratorOrIterable: IterableIterator<T>,
+  length: number
+): Array<T | undefined> => {
+  const iterator = iteratorOrIterable[Symbol.iterator]();
+  return Array.from({ length }, () => iterator.next().value);
+};
+
+const pageSize = 100;
+
+const cacheStore = getGlobalStore<
+  Record<string, [PatientData, number, string]>
+>("patientResultsDataCache", {});
+
+/**
+ * Adds given user testresults data to a cache
+ *
+ * @param patientUuid
+ * @param data {PatientData}
+ * @param indicator UUID of the newest observation
+ */
+export const addUserDataToCache = (
+  patientUuid,
+  data: PatientData,
+  indicator
+) => {
+  cacheStore.setState({ [patientUuid]: [data, Date.now(), indicator] });
+  const currentStateEntries = Object.entries(cacheStore.getState());
+
+  if (currentStateEntries.length > 3) {
+    currentStateEntries.sort(([, [, dateA]], [, [, dateB]]) => dateB - dateA);
+    cacheStore.setState(Object.fromEntries(currentStateEntries.slice(0, 3)));
+  }
+};
+
+const getLatestObsUuid = async patientUuid => {
+  const request = fhirObservationRequests({
+    patient: patientUuid,
+    category: "laboratory",
+    _sort: "-_date",
+    _summary: "data",
+    _format: "json",
+    _count: "1"
+  });
+
+  return (await request.next().value)?.entry?.[0]?.resource?.id;
+};
+
+/**
+ * Retrieves cached user testresults data
+ * Checks the indicator against the backend while doing so
+ *
+ * @param { string } patientUuid
+ * @param { PatientData } data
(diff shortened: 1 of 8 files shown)


src/widgets/test-results/loadPatientTestData/helpers.ts
```diff
@@ -0,0 +1,250 @@
...
+      Math.ceil(total / pageSize) - CHUNK_PREFETCH_COUNT;
+    responses = [
+      ...responses,
+      ...(await Promise.all(
+        retrieveFromIterator(requests, missingRequestsCount)
+      ))
+    ];
+  }
+
+  console.log({ responses });
```
FlorianRappl: Do we need this log output?

src/widgets/test-results/loadPatientTestData/helpers.ts
```diff
@@ -0,0 +1,250 @@
...
+  if (!!data && (await getLatestObsUuid(patientUuid)) === indicator)
+    return data;
+};
+
+/**
+ * Iterator
+ * @param queries
+ */
+function* fhirObservationRequests(queries: Record<string, string>) {
+  const fhirPathname = "/openmrs/ws/fhir2/R4/Observation";
```
FlorianRappl: I think the leading `/openmrs` is actually a variable. It may be different per configuration. At least that's the purpose of `window.openmrsBase` afaik.

src/widgets/test-results/loadPatientTestData/helpers.ts
```diff
@@ -0,0 +1,250 @@
...
+ */
+export const loadPresentConcepts = (
+  entries: ObsRecord[]
+): Promise<ConceptRecord[]> =>
+  Promise.all(
+    [...new Set(entries.map(getEntryConceptClassUuid))].map(
+      conceptUuid =>
+        conceptCache[conceptUuid] ||
+        (conceptCache[conceptUuid] = fetch(
+          "/openmrs/ws/rest/v1/concept/" + conceptUuid + "?v=full"
```
FlorianRappl: Same as above - `/openmrs` is configurable.

FlorianRappl: Can we also prefer to use template strings? Would be splendid.

src/widgets/test-results/loadPatientTestData/helpers.ts
```diff
@@ -0,0 +1,250 @@
...
+        ).then(res => res.json()))
+    )
+  );
+
+/**
+ * returns true if no value is null or undefined
```
FlorianRappl, line 157: Why do we need to know? I think thats where `??` is great.

src/widgets/test-results/overview/helpers.tsx
```diff
@@ -0,0 +1,80 @@
+import React from "react";
+import styles from "./lab-results.scss";
+import { Information16 } from "@carbon/icons-react";
```
FlorianRappl: For the icons (and the components) - please use submodule imports as shown in one of the recent standups.

--> I think in this codebase we also exclusively should use submodule imports, i.e., imports like `import Information16 from '@carbon/icons-react/es/components/information/16'` or so.

src/widgets/test-results/overview/helpers.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { TableRow } from "carbon-components-react";
+import { OBSERVATION_INTERPRETATION } from "../loadPatientTestData/helpers";
+
+export function formatDate(date: Date) {
```
FlorianRappl, line 8: I think `dayjs` does just that.

src/widgets/test-results/overview/useOverviewData.ts
```diff
@@ -0,0 +1,54 @@
...
+  const { sortedObs, loaded, error } = usePatientResultsData(patientUuid);
+  const [overviewData, setDisplayData] = React.useState([]);
+
+  React.useEffect(() => {
+    setDisplayData(
+      Object.entries(sortedObs)
+        .map(([panelName, { entries, type, uuid }]) => {
+          const newestEntry = entries[0];
+
+          let data;
```
FlorianRappl: Type?

FlorianRappl (commented):
I think we are on the right track here! :beers:

src/widgets/test-results/loadPatientTestData/helpers.ts
nmalyschkin: Would you say, it is okay to use nullish coalescing this way?
```ts
if(value ?? false) {
// ...
}
```

FlorianRappl: Yes, I'd say so

FlorianRappl (approved):

gracepotma:
Hooray!! Thank you @nmalyschkin!! Great to see this making it's way into the repo here.

PR #208, opened 2021-03-10
integrate test-results recent-overview as a widget into patient-chart

nmalyschkin:


src/index.ts
```diff
@@ -81,6 +81,18 @@ function setupOpenMRS() {
       }
     ],
     extensions: [
+      {
+        id: "test-results-widget",
+        slot: "test-results-widget",
```
FlorianRappl: ID and Slot should be different.

The slot should be suffixed with `-slot`, or a given slot should be chosen. In this guess you want to have this part in the summary dashboard? Choose `patient-chart-summary-dashboard-slot` as slot.

src/widgets/test-results/overview/recent-overview.component.tsx
```diff
@@ -0,0 +1,160 @@
+import React from "react";
+
+import { Table16, ChartLine16 } from "@carbon/icons-react";
```
FlorianRappl: Import from the respective submodules.

src/widgets/test-results/overview/recent-overview.component.tsx
```diff
@@ -0,0 +1,160 @@
...
+  Table,
+  TableContainer,
+  TableHead,
+  TableHeader,
+  TableRow,
+  TableCell,
+  TableBody,
+  TableToolbarContent,
+  TableToolbar
+} from "carbon-components-react";
```
FlorianRappl: Import from the respective submodules.

src/widgets/test-results/overview/recent-overview.component.tsx
```diff
@@ -0,0 +1,160 @@
...
+import { useHistory } from "react-router";
```
FlorianRappl: `react-router-dom`

FlorianRappl: (And yes, even though `react-router-dom` includes `react-router` note that in general this *may* make a difference. We only share `react-router-dom` (https://github.com/openmrs/openmrs-esm-core/blob/master/packages/esm-app-shell/src/dependencies.ts#L6) which means `react-router` would be bundled in and result in weird / wrong behavior...

FlorianRappl (commented):
Great work! :beers:

src/index.ts
nmalyschkin: `patient-chart-summary-dashboard-slot` does not exist as a slot. All other widgets use a 1-to-1 mapping with their defined slots. Is this going to change in the near future?

FlorianRappl: > patient-chart-summary-dashboard-slot does not exist as a slot.

If slots exist or don't exist does not make a difference. You also don't need to attach to a slot immediately.

> All other widgets use a 1-to-1 mapping with their defined slots.

This is / was just a temporary workaround to get them in as extensions.

> Is this going to change in the near future?

Yes - I think this was announced the last two weekly dev standups. We had the longer session on Monday (sorry during your vacation) where this system was proposed in greater depth. I think the video recording for it was shared already by Grace.

FlorianRappl:
Cannot review when checks are failing. Are you committing with `--no-verify`? Please enable the hooks to get alarmed before pushing.

> /home/runner/work/openmrs-esm-patient-chart-widgets/openmrs-esm-patient-chart-widgets/src/widgets/test-results/overview/overview.component.tsx
> Error:   40:3  error  Unexpected console statement  no-console

nmalyschkin:
> Cannot review when checks are failing. Are you committing with `--no-verify`? Please enable the hooks to get alarmed before pushing.

My bad, my `yarn link` was messing with the type checking, witch swallowed the failing check. It's fixed now.

FlorianRappl (approved):
LGTM - thanks :beers:!

PR #211, opened 2021-03-15
MF-180: Use esm-framework/mock as base mock

samuelmale:
Ticket: https://issues.openmrs.org/browse/MF-180


jest.config.json
```diff
@@ -8,6 +8,7 @@
     "\\.(jpg|jpeg|png)$": "<rootDir>/__mocks__/file.mock.ts",
     "^@carbon/icons-react/es/(.*)$": "@carbon/icons-react/lib/$1",
     "^carbon-components-react/es/(.*)$": "carbon-components-react/lib/$1",
+    "@openmrs/esm-framework/mock": "@openmrs/esm-framework/mock",
```
FlorianRappl: What is this?

What should happen is that

``` 
"@openmrs/esm-framework": "@openmrs/esm-framework/mock"
```

replaces the old entry.

The old entry should be completely removed. If the `useCurrentPatient` should be overriden or removed this can be done in the setup, or in each relevant test module (would be better in any case, too, as it makes this more transparent).

samuelmale: Sure, I thought using global overrides was more ideal since refactoring per test would introduce a lot of "common code" (boilerplate code) across the test suites. 

> (would be better in any case, too, as it makes this more transparent).

Like you said it would be better, let me refactor this :)

FlorianRappl: Yeah if we want to have it global this would work, but then using the test setup would work, too.

I guess at the end its a bit about taste here - but I would love to have it at least in blueprint in an idealized form.

samuelmale: I have come to believe this would make more sense to the other side of the world like java:

```java

public abstract class ChartWidgetsMocks extends BaseMocks {
     // Add all your overrides in one place
}

```

I think it's easier to maintain these mocks, if a given test case expects a different behaviour, then it should handle this at the setup level?

FlorianRappl: Hm not sure I understand. The class example confuses me, too.

What is your specific question?

brandones: Right, @samuelmale , what Florian is suggesting (I think) is that instead of overriding the mock in a local mock file, which is what you've done, you should override the mock in `src/setupTests.ts`. You can do something like

```typescript
import { useCurrentPatient } from "@openmrs/esm-framework/mock"

useCurrentPatient.mockReturnValue([false, mockPatient, mockPatient.id, null]);
```

(assuming that `useCurrentPatient` is already a `jest.Mock`.)

samuelmale: Thanks for the insight, updated.

src/widgets/allergies/allergy-form.test.tsx
```diff
@@ -9,7 +9,7 @@ import {
   mockSaveAllergyResponse,
   mockUpdatedAllergyResult
 } from "../../../__mocks__/allergies.mock";
-import { mockPatientId } from "../../../__mocks__/openmrs-esm-framework.mock";
+import { mockPatientId } from "../../setupTests";
```
FlorianRappl: Ouch... No!

This file is used to initialize the tests. It should not export anything. It should not be used in imports. Its run by Jest implicitly.

If you have the structure we see above, its an indicator that you should not import anything, but declare that explicitly in the code.

src/widgets/conditions/conditions-form.test.tsx
```diff
@@ -6,7 +6,7 @@ import {
   createPatientCondition,
   updatePatientCondition
 } from "./conditions.resource";
-import { mockPatientId } from "../../../__mocks__/openmrs-esm-framework.mock";
+import { mockPatientId } from "../../setupTests";
```
FlorianRappl: Declare explicitly.

src/widgets/immunizations/immunizations-form.test.tsx
```diff
@@ -8,7 +8,7 @@ import { ImmunizationsForm } from "./immunizations-form.component";
 import { savePatientImmunization } from "./immunizations.resource";
 import { getStartedVisit, visitItem } from "../visit/visit-utils";
 import { mockSessionDataResponse } from "../../../__mocks__/session.mock";
-import { mockPatientId } from "../../../__mocks__/openmrs-esm-framework.mock";
+import { mockPatientId } from "../../setupTests";
```
FlorianRappl: Declare explicitly.

src/widgets/allergies/allergy-form.test.tsx
FlorianRappl: (You can also just do:

```diff
-import { mockPatientId } from "../../setupTests";
+import { mockPatient } from "../../../__mocks__/patient.mock";
+const mockPatientId = mockPatient.id;
```

)

src/setupTests.ts
```diff
@@ -1 +1,9 @@
 import "@testing-library/jest-dom/extend-expect";
+import { useCurrentPatient, useConfig } from "@openmrs/esm-framework/mock";
+import { mockPatient } from "../__mocks__/patient.mock";
+import { ConfigMock } from "../__mocks__/chart-widgets-config.mock";
+
+useCurrentPatient.mockReturnValue([false, mockPatient, mockPatient.id, null]);
+useConfig.mockReturnValue(ConfigMock);
+
+export const mockPatientId = mockPatient.id;
```
FlorianRappl: Remove this line

FlorianRappl (changes requested):

FlorianRappl (approved):
LGTM - thanks! :beers:

PR #217, opened 2021-03-16
MF-494 : Introduce search filter in the All page of the forms widget

donaldkibet:
### What does this PR do?

1. Add ability to filters forms

### Screenshoot

screencast 2021-03-16 16-18-39


__mocks__/openmrs-esm-framework.mock.tsx
```diff
@@ -53,3 +57,5 @@ export const mockPatientId = mockPatient.id;
 export const showToast = jest.fn();
 
 export const ExtensionSlot = ({ children }) => <>{children}</>;
+
+export const navigate = jest.fn();
```
brandones: At some point let's delete this file and replace the line in the Jest config with `@openmrs/esm-framework/mock`. No need to include it in this PR if it doesn't work right off the bat.

src/widgets/forms/form-view.component.tsx
```diff
@@ -23,8 +30,23 @@ const FormView: React.FC<FormViewProps> = ({
   patientUuid,
   encounterUuid
 }) => {
+  const searchTimeout = 300;
```
brandones: Nit, and matter of opinion: unless this is used in multiple places, there's not much of a point to not just inline it. The additional declaration here at the top just adds complexity.

src/widgets/forms/form-view.component.tsx
```diff
@@ -23,8 +30,23 @@ const FormView: React.FC<FormViewProps> = ({
...
   const { t } = useTranslation();
   const [activeVisit, setActiveVisit] = React.useState<visitItem>();
+  const [searchTerm, setSearchTerm] = React.useState<string>(null);
+  const [allForms, setAllForms] = React.useState<Array<Form>>(forms);
```
brandones, line 41: `allForms` doesn't seem like a very good name, since `forms` is in fact all forms, and `allForms` is a subset of those forms. Maybe `visibleForms`?

src/widgets/forms/forms-utils.ts
```diff
@@ -28,3 +28,9 @@ export function filterAvailableAndCompletedForms(
 export const areFormsEqual = (a: Form, b: Form): boolean => {
   return a !== null && b !== null && a.uuid === b.uuid;
 };
+
+export const filterFormsByName = (formName: string, forms: Array<Form>) => {
+  return forms.filter(
+    form => form.name.toLowerCase().search(formName.toLowerCase()) !== -1
+  );
+};
```
brandones: Why not just include this function in the component? It's only used in that one place. This just adds a layer of indirection.

src/widgets/forms/form-view.component.tsx
```diff
@@ -53,17 +75,50 @@ const FormView: React.FC<FormViewProps> = ({
...
+          <p className={styles.formResultsLabel}>
+            {allForms.length} {t("matchFound", "match found")}
+          </p>
+        )}
+        {isEmpty(allForms) && !isEmpty(searchTerm) && (
+          <Tile light className={styles.formTile}>
+            <EmptyDataIllustration />
+            <p className={styles.content}>
+              {t(
+                "sorryNoFormsHaveBeenFound",
```
brandones:
```suggestion
                "noFormsFound",
```

src/widgets/forms/form-view.component.tsx
```diff
@@ -53,17 +75,50 @@ const FormView: React.FC<FormViewProps> = ({
...
+                "Sorry, no forms have been found"
+              )}
+            </p>
+            <p className={styles.action}>
+              {t(
+                "trySearchingForTheFormUsingAnAlternativeNameorKeyword",
```
brandones:
```suggestion
                "formSearchHint",
```

brandones: (your names were good, don't worry about this kind of thing in the future. This is a very minor improvement and coming up with good names for things is often very hard)

src/widgets/forms/forms.component.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { fetchAllForms, fetchPatientEncounters } from "./forms.resource";
+import { of } from "rxjs";
+import {
+  mockForms,
+  mockPatientEncounters
+} from "../../../__mocks__/forms.mock";
+
+const mockNavigate = navigate as jest.Mock;
+const mockFetchAllForms = fetchAllForms as jest.Mock;
+const mockFetchPatientEncounters = fetchPatientEncounters as jest.Mock;
```
brandones, line 15: Matter of opinion, but this is data specific to the forms test, so I don't know why it doesn't just live in the forms test.

src/widgets/forms/forms.component.tsx
```diff
@@ -68,6 +73,20 @@ const Forms: React.FC<any> = () => {
...
+  React.useEffect(() => {
+    const filledForms = allForms.map(form => {
+      let checkedForm: Form;
+      completedForms.map(completeForm => {
+        if (completeForm.uuid === form.uuid) {
+          completeForm.checked = true;
+          checkedForm = completeForm;
+        }
+      });
+      return checkedForm ? checkedForm : form;
```
brandones: Why not just

```suggestion
      completedForms.map(completeForm => {
        if (completeForm.uuid === form.uuid) {
          form.checked = true;
        }
      });
      return form;
```

src/widgets/forms/forms.component.tsx
```diff
@@ -68,6 +73,20 @@ const Forms: React.FC<any> = () => {
...
+    });
+    setFilledForms(filledForms);
+  }, [allForms, completedForms]);
```
brandones: Wait, how is `filledForms` different from `completedForms`?

src/widgets/types.ts
```diff
@@ -374,6 +374,7 @@ export type Form = {
   retired: boolean;
   encounterTypeUuid?: string;
   encounterTypeName?: string;
+  checked?: boolean;
```
brandones: Hmm... I think `checked` isn't really a property a `Form`. It's a characteristic of how we display the form (think "V" in "MVC"). If that's information that the view can infer on its own, then just have the view do so. If it actually expresses something deeper about the Form, e.g. that the form is complete, give it a name that signifies that, like `complete?: boolean;`.

brandones (changes requested):
Very nice work, @donaldkibet , but a few things need a little refining. It shouldn't be hard given that it looks like you wrote some really excellent tests.

src/widgets/forms/forms.component.tsx
brandones: Oh, I see, the `form` and the `completedForm` objects may be different? It's a little confusing that we have multiple copies of forms floating around in this component. But don't worry about it, this is probably fine.

src/widgets/forms/forms.component.tsx
donaldkibet: `completedForms` are forms a user has filled for a patient, these forms are displayed on the second tab `Complete`.
`filledForms` are basically all forms but have the `Form.completed` property updated to display it as already completed under the `All` tab on the content switcher

src/widgets/forms/form-view.component.tsx
donaldkibet: Update the name to `forms`

src/widgets/forms/forms.component.test.tsx
donaldkibet: am confused here, this is `forms.component.test.tsx` file that contains the tests for `forms.component.tsx` maybe you can point me to the best place to have the data specified

brandones: All I mean is that I don't think `forms.mock.ts` and `visits.mock.ts` really need to exist as their own files — they can just be inlined here at the top of the test file. It's not important though. A matter of opinion.

src/widgets/forms/forms.component.tsx
brandones: Why not just add the `completed` property to `forms` and get rid of `filledForms` entirely?

And then why not get rid of the `completedForms` state variable and just do `const completedForms = forms.filter(f => f.completed)`?

brandones (approved):
Awesome job. I think the way forms are modeled and named in these components can still be improved, but it's not really what this PR is about. Glad to see this go in.

FlorianRappl (approved):
:beers:

PR #220, opened 2021-03-19
Add forms pagination and see all link

donaldkibet:
### What does this PR do?

1. Adds forms-pagination
2. Ability to navigate away when `See all` link is clicked
3. Add patientChart pagination.

### To od
- [ ] fix broken test because the mocked [openmrs-esm-framework/mock](https://github.com/openmrs/openmrs-esm-core/blob/master/packages/esm-framework/mock.tsx) interpolateString function returns `undefined` resulting in the test to fail

### Screenshoot
screencast 2021-03-19 16-20-22

### DIFF
--- a/src/ui-components/pagination/pagination.component.tsx
+++ b/src/ui-components/pagination/pagination.component.tsx
@@ -1,0 +1,76 @@
+import React from "react";
+import { useTranslation } from "react-i18next";
+import styles from "./pagination.component.scss";
+import Pagination from "carbon-components-react/es/components/Pagination";
+import { ConfigurableLink, useCurrentPatient } from "@openmrs/esm-framework";
+import isEmpty from "lodash-es/isEmpty";
+
+interface PatientChartPaginationProps {
+  items: Array<unknown>;
+  pageNumber: number;
+  pageSize: number;
+  onPageNumberChange?: any;
+  pageUrl: string;
+}
+
+/**
+ * @param items The items to be paged.
+ * @param pageNumber The number of the page to be returned as the first result element. Not zero-based!
+ *                   The first page has the number 1.
+ * @param pageSize The number of items per page.
+ * @param onPageNumberChange The function called when page Number is changed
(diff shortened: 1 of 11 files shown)


src/ui-components/pagination/pagination.component.tsx
```diff
@@ -0,0 +1,76 @@
...
+  onPageNumberChange,
+  pageNumber,
+  pageUrl
+}) => {
+  const { t } = useTranslation();
+  const [, , patientUuid] = useCurrentPatient();
+  const chartBasePath = "${openmrsSpaBase}/patient/:patientUuid/chart/".replace(
+    ":patientUuid",
+    patientUuid
+  );
```
brandones:
```suggestion
  const chartBasePath = `${openmrsSpaBase}/patient/${patientUuid}/chart/`;
```

src/ui-components/pagination/pagination.component.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  const generatePageSizes = () => {
+    const numberOfPages = Math.ceil(items?.length / pageSize);
+    return [...Array(numberOfPages).keys()].map(x => {
+      return ++x * pageSize;
```
brandones:
```suggestion
      return (x + 1) * pageSize;
```

Makes my brain hurt less

src/ui-components/pagination/pagination.component.tsx
```diff
@@ -0,0 +1,76 @@
...
+    });
+  };
+
+  return (
+    <>
+      {items.length > 0 && (
```
brandones, line 45: I might be wrong, but I think React will get mad if you try to render `false`. I think you want `items.length ? whatever : null`.

donaldkibet: When using `Logical && Operator` if the condition is true, the element right after `&&` will appear in the output. If it is false, React will ignore and skip it.

brandones (approved):
LGTM, once failing checks are fixed

PR #221, opened 2021-03-21
Corrected the URL to Vitals Form

ujjwaltyagi355:
This is to solve the issues [MF-499](https://issues.openmrs.org/browse/MF-499).
The record vitals page was not loading due to some error in the URL.
Made changes to the URL and now the page is loading as described in the wireframe of the issue.

image

Thank you.

### DIFF
--- a/src/widgets/vitals/vitals-overview.component.tsx
+++ b/src/widgets/vitals/vitals-overview.component.tsx
@@ -111,7 +111,7 @@ const VitalsOverview: React.FC<VitalsOverviewProps> = () => {
     });
 
   const launchVitalsBiometricsForm = () => {
-    const url = `/patient/${patientUuid}/vitalsbiometrics/form`;
+    const url = "/patient/:patientUuid/vitalsbiometrics/form";
     switchTo("workspace", url, {
       title: t("recordVitalsAndBiometrics", "Record Vitals and Biometrics")
     });


src/widgets/vitals/vitals-overview.component.tsx
```diff
@@ -111,7 +111,7 @@ const VitalsOverview: React.FC<VitalsOverviewProps> = () => {
     });
 
   const launchVitalsBiometricsForm = () => {
-    const url = `/patient/${patientUuid}/vitalsbiometrics/form`;
+    const url = "/patient/:patientUuid/vitalsbiometrics/form";
```
FlorianRappl, line 114: This is an invalid URL.

Should not be the path template, but the actual path.

FlorianRappl (changes requested):

src/widgets/vitals/vitals-overview.component.tsx
ujjwaltyagi355: Hello, @FlorianRappl, 
Just clarifying:
path like : "${openmrsSpaBase}/patient/:patientUuid/vitalsbiometrics/form"

FlorianRappl: No.

```ts
`/patient/${patientUuid}/vitalsbiometrics/form`
```

is correct. In your screenshot also all vitals are empty, i.e., no valid patient selected. Make sure the patientUuid is defined and valid though.

ujjwaltyagi355: Hello @FlorianRappl,
> Make sure the patientUuid is defined and valid

image

I think it's defined correctly here. Because I made a few code changes to check for its value:

image



> In your screenshot also all vitals are empty, i.e., no valid patient selected.

This page is to add the vitals and Biometrics. 

I tried to load the form with the help of the patientUuid from the chart/summary URL. The same page gets loaded.

image

And I was able to add the vitals to the patient as well through the page that gets rendered.

image

Do have a look, please.
Thank you.

FlorianRappl: Are you making the requested changes now?

Because I don't see any push. Still it's using the invalid format.

ujjwaltyagi355: @FlorianRappl,
I am working on it.
`/patient/${patientUuid}/vitalsbiometrics/form`
This URL is not working, even though the `patientUuid` is correct. 
I just wanted to bring to your notice that the above changes that I made are working fine. 
I will soon commit the changes as per the mentioned guideline.
Thank you.

FlorianRappl: I'm not sure here.

Because we use the same approach in the drug ordering and it works. Also there is not any part that does such replacement that I am aware of.

> I will soon commit the changes as per the mentioned guideline.

What changes? I guess there may be a misunderstanding here. You mean the requested changes?

ujjwaltyagi355: @FlorianRappl,
When we click on the "Add" button on the vitals widget, it takes us to an empty page.

vitals

For this only the issue [MF-499](https://issues.openmrs.org/browse/MF-499), is created.
And I am trying to work on the link.
Do tell me if I am missing something here.
Thank you.

FlorianRappl: Clicking on the plus button triggers (among other things):

```
https://openmrs-spa.org/openmrs/ws/rest/v1/visit?patient=null&v=custom:(uuid,encounters:(uuid,encounterDatetime,form:(uuid,name),location:ref,encounterType:ref,encounterProviders:(uuid,display,provider:(uuid,display))),patient:(uuid,uuid),visitType:(uuid,name,display),attributes:(uuid,display,value),location:(uuid,name,display),startDatetime,stopDatetime)
```

Note that `patient` is `null`. This problem is **not** exclusive to vitals, but can also be seen on other (visit-related) parts. I guess it may have to do with usage of React Router instead of vanilla JS URL properties. But that is just a guess. Actually finding the root cause and solving this is what should be done here.

FlorianRappl: Note that the problem above is presumably not the root cause, but just some side-effect of a bigger issue. Most requests are fine and contain the valid patient UUID - I guess there is some issue in the code with the response.

But again, just observations. The bottom line is that there is no issue with the current URL, but rather with the handling of the response / the component itself.

ujjwaltyagi355: @FlorianRappl,
What do you suggest should be the further steps to correct this?

Bawanthathilan, line 114: this path in incorrrect

Bawanthathilan (changes requested):

PR #224, opened 2021-03-23
Implement see all link and widget pagination

donaldkibet:
### What does this PR do

- [x] Implement see all link for widgets
- [x] Implement pagination for all widgets 

### Screenshoot
screencast 2021-03-23 17-18-01

### DIFF
--- a/src/widgets/allergies/allergies-overview.component.tsx
+++ b/src/widgets/allergies/allergies-overview.component.tsx
@@ -23,9 +23,8 @@ import AllergyForm from "./allergy-form.component";
 import EmptyState from "../../ui-components/empty-state/empty-state.component";
 import ErrorState from "../../ui-components/error-state/error-state.component";
 import styles from "./allergies-overview.scss";
-import isEmpty from "lodash-es/isEmpty";
-import paginate from "../../utils/paginate";
 import PatientChartPagination from "../../ui-components/pagination/pagination.component";
+import { usePaginate } from "../../utils/use-paginate";
 
 const AllergiesOverview: React.FC<AllergiesOverviewProps> = () => {
   const allergiesToShowCount = 5;
@@ -33,12 +32,9 @@ const AllergiesOverview: React.FC<AllergiesOverviewProps> = () => {
   const [isLoadingPatient, patient] = useCurrentPatient();
   const [allergies, setAllergies] = React.useState<Array<Allergy>>(null);
   const [error, setError] = React.useState(null);
-  const [showAllAllergies, setShowAllAllergies] = React.useState(false);
   const displayText = t("allergyIntolerances", "allergy intolerances");
   const headerTitle = t("allergies", "Allergies");
   const [pageNumber, setPageNumber] = React.useState(1);
-  const [pageSize, setPageSize] = React.useState(5);
-  const [currentPage, setCurrentPage] = React.useState([]);
 
   React.useEffect(() => {
     if (!isLoadingPatient && patient) {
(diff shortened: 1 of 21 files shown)


src/widgets/allergies/allergies-overview.component.tsx
```diff
@@ -71,22 +77,31 @@ const AllergiesOverview: React.FC<AllergiesOverviewProps> = () => {
     openWorkspaceTab(AllergyForm, t("allergiesForm", "Allergies Form"));
   };
 
+  const handlePageChange = ({ page }) => {
+    setPageNumber(page);
+  };
+
+  React.useEffect(() => {
+    if (!isEmpty(allergies)) {
```
FlorianRappl: Hm to me all of that rather sounds like a use of `useMemo` than `useEffect`. You can just *always* compute the value for the current page entries, so `useMemo` may be great since it improves perf.

src/widgets/appointments/appointments-overview.component.tsx
```diff
@@ -34,6 +37,24 @@ const AppointmentsOverview: React.FC<AppointmentOverviewProps> = () => {
   const startDate = dayjs().format();
   const displayText = t("appointments", "appointments");
   const headerTitle = t("appointments", "Appointments");
+  const [pageNumber, setPageNumber] = React.useState(1);
+  const [pageSize, setPageSize] = React.useState(5);
+  const [currentPage, setCurrentPage] = React.useState([]);
```
FlorianRappl: Same as above. I'd rather use `useMemo`. If you do that you could also easily introduce a hook like `usePaginate`.

src/widgets/allergies/allergies-overview.component.tsx
seankwon: i think if this were using use memo, it would be implemented like so.

```
const [page, allPages] = React.useMemo(() => {
   return paginate<any>(allergies, pageNumber, pageSize);
}, [!isEmpty(allergies), pageNumber, pageSize]]);
```
if you do it this way, you probably don't need initialize state for page / currentpage, right? (ps, i havent tried this code yet.)

src/widgets/vitals/vitals-overview.component.tsx
```diff
@@ -158,61 +167,55 @@ const VitalsOverview: React.FC<VitalsOverviewProps> = () => {
...
+                  isSortable={true}
+                  size="short"
+                >
+                  {({ rows, headers, getHeaderProps, getTableProps }) => (
+                    <Table {...getTableProps()}>
+                      <TableHead>
+                        <TableRow>
+                          {headers.map(header => (
+                            <TableHeader
+                              className={`${styles.productiveHeading01} ${styles.text02}`}
```
seankwon, line 177: i think this also needs a key?

src/widgets/vitals/vitals-overview.component.tsx
```diff
@@ -84,6 +84,13 @@ const VitalsOverview: React.FC<VitalsOverviewProps> = () => {
     }
   }, [patientUuid]);
 
+  React.useEffect(() => {
+    if (!isEmpty(vitals)) {
+      const [page, allPages] = paginate<any>(vitals, pageNumber, pageSize);
```
seankwon: yea, same here - i think if you implement this with useMemo, i'm not sure you would need state for currentPage / setCurrentPage

src/widgets/biometrics/biometrics-overview.component.tsx
```diff
@@ -60,34 +64,47 @@ const BiometricsOverview: React.FC<BiometricsOverviewProps> = () => {
...
       return () => sub.unsubscribe();
     }
   }, [patientUuid, config.concepts.weightUuid, config.concepts.heightUuid]);
 
+  const handlePageChange = ({ page }) => {
+    setPageNumber(page);
+  };
+
+  React.useEffect(() => {
+    if (!isEmpty(biometrics)) {
```
seankwon: same here, you could use useMemo as well

FlorianRappl:
Also what's the relation to #213, #214, and #215 @donaldkibet ?

donaldkibet:
> Also what's the relation to #213, #214, and #215 @donaldkibet ?

There is an overlap with the PRs on this [ticket](https://issues.openmrs.org/browse/MF-439)

PR #227, opened 2021-03-24
Fix patient banner overflow menu margin

denniskigen:
Applies a bottom margin to the overflow menu to keep it in line with the rest of the items in the patient banner. See before and after shots below for context.

Screenshot 2021-03-24 at 16 47 03

Screenshot 2021-03-24 at 16 51 43


FlorianRappl (approved):
I don't like negative margins, but sometimes they are useful.

PR #231, opened 2021-03-29
Use an SVG placeholder as default patient photo

denniskigen:
Changes the default patient placeholder to an SVG instead of the PNG file we had before. (Cheers Ciaran for the handy SVG)

SVG's offer myriad advantages on the web. See https://css-tricks.com/using-svg/ and https://vecta.io/blog/comparing-svg-and-png-file-sizes.

Screenshot 2021-03-29 at 22 19 40

Related: https://github.com/openmrs/openmrs-esm-patient-registration/pull/87.


src/assets/placeholder.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="80" height="80" xmlns="http://www.w3.org/2000/svg">
```
HerbertYiga, line 1: on introducing the  placeholder.svg ,dont we need to drop the  placeholder.png

FlorianRappl (approved):

PR #232, opened 2021-03-30
MF-534: Fix vitals signs display values when units are absent

denniskigen:
[Ticket](https://issues.openmrs.org/browse/MF-524)

- Change the display logic of the vital signs table headers so that units are shown only when present.
- Change the display logic of the vital signs radio buttons in the chart view so that units are shown only when present.
- Change the `Rate` label to `R. Rate` per the designs.
- Add tests 🚀

Fixes:
Screenshot 2021-03-30 at 10 31 41
Screenshot 2021-03-30 at 14 05 22

To:
Screenshot 2021-03-30 at 14 08 58
Screenshot 2021-03-30 at 14 09 07

### DIFF
--- a/src/widgets/vitals/vitals-chart.component.tsx
+++ b/src/widgets/vitals/vitals-chart.component.tsx
@@ -88,27 +88,27 @@ const VitalsChart: React.FC<VitalsChartProps> = ({
   const vitalSigns = [
     {
       id: "bloodPressure",
-      title: `BP (${bloodPressureUnit})`,
+      title: `BP ${bloodPressureUnit ? `(${bloodPressureUnit})` : ""}`,
       value: "systolic"
     },
     {
       id: "oxygenSaturation",
-      title: `SPO2 (${oxygenSaturationUnit})`,
+      title: `SPO2 ${oxygenSaturationUnit ? `(${oxygenSaturationUnit})` : ""}`,
       value: "oxygenSaturation"
     },
     {
       id: "temperature",
-      title: `Temp (${temperatureUnit})`,
+      title: `Temp ${temperatureUnit ? `(${temperatureUnit})` : ""}`,
       value: "temperature"
     },
     {
       id: "Respiratory Rate",
-      title: `R.Rate ${respiratoryRateUnit}`,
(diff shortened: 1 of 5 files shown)


src/widgets/vitals/vitals-chart.component.tsx
```diff
@@ -88,27 +88,27 @@ const VitalsChart: React.FC<VitalsChartProps> = ({
   const vitalSigns = [
     {
       id: "bloodPressure",
-      title: `BP (${bloodPressureUnit})`,
+      title: `BP ${bloodPressureUnit ? `(${bloodPressureUnit})` : ""}`,
```
FlorianRappl: You could also use `${bloodPressureUnit ?? ''}`

denniskigen: Would `??` work too in the case where if `bloodPressureUnit` is truth, I want it to return the unit encased in brackets and an empty string otherwise?

`BP (mmHg)` if units specified
`BP` otherwise

FlorianRappl: Nope, `??` only works for `undefined` and `false`. But I guess that is exactly what you are after (see screenshot).

Needless to say there is the `()` that I did not notice. So scratch that idea. All good (even though a helper would still make sense).

src/widgets/vitals/vitals-chart.component.tsx
```diff
@@ -88,27 +88,27 @@ const VitalsChart: React.FC<VitalsChartProps> = ({
...
       value: "systolic"
     },
     {
       id: "oxygenSaturation",
-      title: `SPO2 (${oxygenSaturationUnit})`,
+      title: `SPO2 ${oxygenSaturationUnit ? `(${oxygenSaturationUnit})` : ""}`,
```
FlorianRappl: Introduce:

```ts
function withUnit(label: string, unit: string | null | undefined) {
  return `${label} ${unit ? `(${unit})` : ''}`;
}
```

you can rewrite these to

```js
title: withUnit('SPO2', oxygenSaturationUnit),
```

denniskigen: Brilliant, thanks @FlorianRappl. Updated.

FlorianRappl (approved):

PR #235, opened 2021-03-31
Post-fix slot to match slot naming convention

donaldkibet:
### What does this PR do?

- [x] Post fix `slot` to extensionSlotName to comply with conventions

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -94,7 +94,7 @@ function setupOpenMRS() {
       },
       {
         id: "test-results-widget",
-        slot: "patient-chart-dashboard-medications",
+        slot: "patient-chart-dashboard-medications-slot",
         load: getAsyncLifecycle(
           () =>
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -152,7 +152,7 @@ function setupOpenMRS() {
         )
       },
       {
-        id: "visit-notes-form-ext",
```
HerbertYiga, line 155: @donaldkibet   do we intend to rename the Id too?

src/index.ts
```diff
@@ -94,7 +94,7 @@ function setupOpenMRS() {
       },
       {
         id: "test-results-widget",
-        slot: "patient-chart-dashboard-medications",
+        slot: "patient-chart-dashboard-medications-slot",
```
FlorianRappl, line 97: Since all of that is essentially outdated and will be replaced tomorrow I would discard the PR.

